PR #261, opened 2018-06-26
Issue 260 - ensure sample project generates trees for all revisions

### LINKED ISSUE #260: Ability to generate trees for all revisions through the client
to fix https://github.com/3drepo/3drepo.io/issues/1042

carmenfan:
This fixes #260 and https://github.com/3drepo/3drepo.io/issues/1042

Problem Pavol had was because we only re-generate the tree for the latest revision (regeneration of tree is needed to ensure the modelID and the teamspace name is correct), meaning the tree is effectively broken if you try to load any older revisions.

This is fixed by ensuring we generate all trees for all revisions when we import the sample project.

### DIFF
--- a/client/src/functions.cpp
+++ b/client/src/functions.cpp
@@ -21,10 +21,12 @@
 
 #include <sstream>
 #include <fstream>
+#include <algorithm>
 
 #include <boost/property_tree/ptree.hpp>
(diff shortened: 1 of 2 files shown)


carmenfan:
I can't reproduce this on dev. Maybe we should take a look at your system lol

Issue #265, opened 2018-10-03
Write UnityAsset to BSON


Charence:
See https://github.com/3drepo/wiki/wiki/Mongo-move-strategy-tasks

PR #272, opened 2018-10-24
ISSUE #265 - Write Unity assets list as BSON

Charence:
Addresses #265 

Write Unity assets list as BSON to DB.

### DIFF
--- a/bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_asset.cpp
+++ b/bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_asset.cpp
@@ -51,7 +51,6 @@ const static std::string MP_LABEL_PROJECT = "model";
 
 AssetModelExport::AssetModelExport(
 	const repo::core::model::RepoScene *scene,
-	repo::core::handler::AbstractDatabaseHandler *handler,
 	const bool vrEnabled
 	) : WebModelExport(scene),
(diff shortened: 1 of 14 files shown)


bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_asset.cpp
```diff
@@ -199,7 +224,24 @@ bool AssetModelExport::generateTreeRepresentation(
 		assetListTree.addToTree(MP_LABEL_DATABASE, scene->getDatabaseName());
 		assetListTree.addToTree(MP_LABEL_PROJECT, scene->getProjectName());
 		jsonTrees[assetListFile] = assetListTree;
+
+		std::string errMsg;
+		if (success &= commitUnityAssets(
```
carmenfan: It doesn't make sense to commit the unit assets here. So after all I don't think we should be passing the handler here.... (sorry 😛)

This is called during initialisAssetBuffer, which is called by the BouncerWrapper to "setup" all the files and info needed to build the bundle, if for some reason this failed, we don't want to get into a situation where we have already committed a part of the work.

I think what's best to do is return it like the `jsonFiles`/`serialisedFaceBuf`, let BouncerWrapper hold it (like everything else it's holding), and pass it back down to bouncer library when it calls `commitAssetBundleBuffers` in `save AssetBundles`

PR #286, opened 2018-12-05
Implemented full import of geometry and materials using Teigha BIM

azhakun:


carmenfan:
@azhakun Can you please make sure this is tested with the gui method I sent via email?

Looking at this it will definitely not work as it doesn't look like it will compile due to the lack of changes on cmake to compile.

The gui method will be able to test all the changes made for geometry, and I can't really construct the cloud solution for you to test properly if it doesn't compile.

PR #321, opened 2019-04-25
Issue 320 - Fileshare support

carmenfan:
This needs the ISSUE_211 in unity repo

- Added gridfsFileHandler and fsFileHandler
- Filehandlers are no longer singletons. Instead, a FileManager is added to manage fileHandlers, and the rest of the library talks to this singleton object.
- FileManager handles the ref node writing instead of FileHandlers
- gridfs writes (apart from the non migrated ones) now go through gridFSFileHandler.
- got rid of the clean functions.

**Config change required**
Bouncer now looks for config.json instead of config.js. The config is now a simple json instead of a javascript file. This is mainly because I can't get C++ to easily read a javascript file. See config_example.json. What you basically have to do is mv config.js to config.json, remove the module.export, make sure everything has quotes around it.

db is now it's own section instead of under bouncer


tools/bouncer_worker/bouncer_worker.js
```diff
@@ -47,6 +47,11 @@
 	let retry = 0;
 	let connClosed = false;
 
+	const fs = require("fs");
+	const configFullPath = path.resolve(__dirname, "config.json");
+	const conf = JSON.parse(fs.readFileSync(configFullPath));
```
Charence: This conflicts with earlier declaration of `conf` on Line 35.

Charence: Shouldn't Line 35 be removed?

carmenfan: > Shouldn't Line 35 be removed?
yes

PR #336, opened 2019-06-11
Issue 335 - Return error and force a failure 

carmenfan:
- We now check if the bad bit is flagged on a ofstream
- return "" (for linkage reference) if it failed
- track and stop the process if it's failed


bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
```diff
@@ -84,11 +86,11 @@ std::string FSFileHandler::uploadFile(
 	const std::vector<uint8_t> &bin
 	)
 {
-	auto hierachy = determineHierachy(keyName);
+
+	auto hierachy = level > 0 ? determineHierachy(keyName) : std::vector<std::string>();
```
Charence: I seem to have problems on OSX when level = 1

Issue #384, opened 2020-09-16
Spawn() doesn't exit sometimes if you don't consume the stdin/stdout buffer


carmenfan:
This is only found to be happening in a specific IFC model, in Ubuntu

Maybe to do with this file is corrupted thus the import is struggling and putting things in stderr, without consuming the data, it seems like it doesn't want to call close, thus the promise never returns
]

By simply adding this: 
![image](https://user-images.githubusercontent.com/11945337/93370337-92a76880-f848-11ea-94b6-1b79f21c6887.png)

It returns fine.

PR #385, opened 2020-09-16
ISSUE #384 add code to consume the std buffers

carmenfan:
@keifgwinn  can you deploy this somewhere and make sure it fixes the problem with the file?


carmenfan:
Verified working by @keifgwinn 

Merging

PR #394, opened 2020-10-22
Issue 393 - Synchro fixes

### LINKED ISSUE #393: Sequences have conflicting hide/show information when imported via linux.
Sequences are not showing as expected on any models imported via linux.

After some investigation, there seems to be IDs that reside in a frame instructed to both hide and show. As show is being called before hide, this is causing it to hide, thus it appears the animation isn't really doing anythig.

Not sure if this is a bouncer problem or a Synchro Reader problem yet.

carmenfan:
This resolves #393 

- ensure the object is removed from the visibility state cache if the visibility resets/changes
- transformation of the final state is applied and assumed to be the default position. An Inverse of the matrix is applied when the transformation resets.

### DIFF
--- a/bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_synchro.cpp
+++ b/bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_synchro.cpp
@@ -215,7 +215,8 @@ void SynchroModelImport::determineMeshesInResources(
 }
 
 repo::core::model::RepoScene* SynchroModelImport::constructScene(
-	std::unordered_map<std::string, std::vector<repo::lib::RepoUUID>> &resourceIDsToSharedIDs
+	std::unordered_map<std::string, std::vector<repo::lib::RepoUUID>> &resourceIDsToSharedIDs,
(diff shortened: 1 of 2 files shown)


bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_synchro.cpp
```diff
@@ -640,32 +679,37 @@ void SynchroModelImport::generateTaskInformation(
...
 		repoInfo << "Getting tasks... ";
 
-		generateTaskInformation(reader->getTasks(), resourceIDsToSharedIDs, scene);
+		auto firstFrame = generateTaskInformation(reader->getTasks(), resourceIDsToSharedIDs, scene);
 
 		repoInfo << "Getting animations... ";
-		auto animInfo = reader->getAnimation();
-
-		auto animation = animInfo.first;
+		auto animation = reader->getAnimation();
```
Charence, line 700: I'm getting `'struct pair<Animation, unordered_map<string, bool>>' has no member named '...'` errors on lines, 712, 737, 767, 770, 780, 800 for things like `name`, `lastFrameVisibility`, etc.

Charence (approved):

Issue #391, opened 2020-10-13
Texture support in BIM files


carmenfan:
#### Preparation
- [x] Get access to some 3D models with textures
- [x] Compile and run the the test program from (Program.cs) the [related ](https://github.com/3drepo/3drepoPlugin/pull/131)branch to get the resultant BIM file
- [x] Examine C# code to understand the BIM format layout 
- [x] *SHOULD*: Create  description/diagram of the BIM format
> See the Notes section below
- [x] Highlight changed areas in the BIM format to allow for textures 
- [x] Pass feedback on for internal inconsistencies in the .NET code for the latest BIM003 format

#### Bouncer dev
- [ ] Complete the notes on getting bouncer dev environment up [here](https://github.com/3drepo/3drepobouncer/issues/396)
- [x] Highlight the key areas that require updating to match the changes in the format
- [x] Write a test to pass on processing the BIM003 file
- [x] Implement code changes
- [x] Probe for texture notes and attempt to binary dump texture to an image file
- [x] Write tests for BIM001/002
- [x] Write edge case tests for the above  
- [x] Write system test for the client exe
- [x] Set up the local environment for converting to unity asset bundles
- [x] Test with large BIM files (via client)
- [x] SHOULD: Export to Asset bundle

### Assumptions
- Changes will be backwards compatibility with v2/1 of the BIM format.

### Questions
- What is sizes for? If its the size of each node, then sizes[0] does make sense...
> Updated format descriptions with correct info about this
- I cannot find a BIM001 label/property in the git history for the exporter was it ever publicly released?
> This is in the format descriptions below
- Can we place small 3d models used for unit testing the .NET code [here](https://github.com/3drepo/tests)?
> Can use the linked repo 

### Observations going through this issue

- metadata is duplicated in the current BIM format. This also complicates importing of the model in bouncer

### Related ticket
see https://github.com/3drepo/3drepoPlugin/issues/105

### Reference 

[Wiki](https://github.com/3drepo/wiki/wiki/BIM-file-format-descriptions) page for the file format descriptions

har00n-haider:
Observations going through this issue:

- metadata is duplicated in the current BIM format. This also complicates importing of the model in bouncer

har00n-haider:
@carmenfan , I'm getting close to testing the solution with with the BIM003 file generated by the plugins. Could you point me to some BIM001/BIM002 files I can use for unit testing?

har00n-haider:
[...]

PR #402, opened 2020-11-12
Issue 391: Adding BIM3 support to bouncer 

har00n-haider:
Implements the BIM file format changes mentioned in #391


test/src/unit/repo_test_utils.h
```diff
@@ -197,3 +199,133 @@ static std::string getRandomString(const uint32_t &iLen)
...
+
+namespace repo
+{
+namespace test
+{
+	class TestLogging
+	{
+	public:
+
+		static void printTestTitleString(std::string title, std::string description)
```
har00n-haider, line 211: @carmenfan, not sure if this is the best place to put this function, would it would be placed alongside the repo errors h file. Or did I just bungle something up and we already had something like this?

har00n-haider:
@carmenfan, ready for review.

Quick summary:

- Updated to BIM3 
- Texture data loads correctly
- Material/Texture node relationships are tied up correctly
- UV channels are being read correctly and rendered
- system and unit test updated for the changes to the BIM3 format  (uv channels added)

test/src/unit/repo_test_utils.h
carmenfan: I don't think we need it really, That's what the 2nd parameter of TEST is for.

If you need it to be more precise then just split them into multiple test cases, with more descriptive test names.

Logging should be kept to minimum and precise ideally. Otherwise it clutters the field and it begins to get extremely hard to read.

test/src/unit/repo_test_utils.h
```diff
@@ -197,3 +199,133 @@ static std::string getRandomString(const uint32_t &iLen)
...
+			repoInfo << "============" << title << "============";
+			repoInfo << description;
+		}
+
+		static void printSubTestTitleString(std::string title)
+		{
+			repoInfo << "------------" << title << "------------";
+		}
+
+		static std::string getStringFromRepoErrorCode(int repoErrorCode)
```
carmenfan, line 222: (just saw this whilst replying to the other question)
@har00n-haider Does this need to exist?

Again, this goes back to the preference of keeping logging precise and minimal. If an error was triggered gtest will tell us what was the expected error code is suppose to be. There's not a lot of need to print this.

This also has a danger of going out of sync with error_codes.h

test/src/unit/repo_test_utils.h
carmenfan: If you really need an explanation of what the test is doing, put it as a code comment.

When you're running the test you don't really care unless it stops working, if it stops working you'd have to refer to the code anyway.

har00n-haider: You know what, that makes total sense, I always used to try and make the test outputs clear as to what they were supposed to be doing, but ultimately no one cares until something goes wrong. I'll refactor and put any useful statements as code comments.

test/src/unit/repo_test_utils.h
har00n-haider: As with above I'll axe this.

har00n-haider:
@carmenfan , have removed the verbosity from the st/ut and the related logging functions.

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -132,25 +137,76 @@ repo::core::model::MaterialNode *RepoModelImport::parseMaterial(const boost::pro
 	materials.insert(materialNode);
 	matNodeList.push_back(materialNode);
 
-	return materialNode;
+	if(textureId >= 0)
+	{
+		textureIdToParents[textureId].push_back(materialNode->getSharedID());
+	}
+	return;
```
carmenfan, line 144: this is not needed?

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -132,25 +137,76 @@ repo::core::model::MaterialNode *RepoModelImport::parseMaterial(const boost::pro
...
+		return;
+	}
+
+	std::string name = fileNameOk ? textureTree.get_child("filename").data() : "";
+	uint32_t byteCount = textureTree.get<uint32_t>("numImageBytes");
+	uint32_t width = textureTree.get<uint32_t>("width");
+	uint32_t height = textureTree.get<uint32_t>("height");
+	uint32_t id = textureTree.get<uint32_t>("id");
+
+	std::vector<uint32_t> DataStartEnd = as_vector<uint32_t>(textureTree, "imageBytes");
```
carmenfan, line 173: use `uint64_t`

This number can overflow on 32bit if the geometry buffer is > 2gb (which does happen sometimes)

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -132,25 +137,76 @@ repo::core::model::MaterialNode *RepoModelImport::parseMaterial(const boost::pro
...
+		textureIdToParents[textureId].push_back(materialNode->getSharedID());
+	}
+	return;
+}
+
+void RepoModelImport::parseTexture(
+	const boost::property_tree::ptree& textureTree, 
+	char * const dataBuffer)
+{
+	bool fileNameOk	 = textureTree.find("filename") != textureTree.not_found();
```
carmenfan, line 151: can we put these strings as constants somewhere? (top of file, top of function, a new file called `repo_model_import_3drepo_constants.h` , whatever makes the most sense)

I know this file itself doesn't seem to do this, but it 'd be a good habit to stop introducing magic strings everywhere, esp we reference it again on a different line

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -132,25 +137,76 @@ repo::core::model::MaterialNode *RepoModelImport::parseMaterial(const boost::pro
...
+			repo::core::model::RepoBSONFactory::makeTextureNode(
+				name,
+				data,
+				byteCount,
+				width,
+				height,
+				REPO_NODE_API_LEVEL_1));
+	
+	repo::core::model::TextureNode* tmpTexture = new repo::core::model::TextureNode();
+	*tmpTexture = textureNode->cloneAndAddParent(textureIdToParents[id]);
```
carmenfan, line 188: alter the `makeTextureNode` api to optionally take in parents array. cloning is expensive so we want to avoid doing that unless it's necessary

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -224,11 +276,24 @@ RepoModelImport::mesh_data_t RepoModelImport::createMeshRecord(
 			}
 		}
 
+		if (props->first == REPO_IMPORT_UV)
```
carmenfan, line 279: Can we rewrite this to a switch case? (I know this isn't your fault, it's pre-established!)

It's a bit stupid that we have all these if statements, if props->first == REPO_IMPORT_UV is true, it's not going to be true for REPO_IMPORT_MATERIAL or any other of the ifs around here.

So it's better if it's if...else, or a switch statement. The latter is cleaner i think

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -261,7 +326,7 @@ RepoModelImport::mesh_data_t RepoModelImport::createMeshRecord(
 		matParents[materialID].push_back(sharedID);
 	}
 
-	mesh_data_t result = { vertices, normals, faces, boundingBox, parentID, sharedID };
+	mesh_data_t result = { parentBimId, vertices, normals, uvChannels, faces, boundingBox, parentID, sharedID };
```
carmenfan, line 329: `parentBimId` is probably just added for debugging? do you still want it in there?

test/src/system/st_3drepobouncerClient.cpp
```diff
@@ -123,6 +123,26 @@ static int runProcess(
...
 }
 
+
+static int testUpload (
+	std::string mongoDbName,
+	std::string projectName,
+	std::string fileName
+	)
+{
+	std::string bimUpload = produceUploadArgs(
```
carmenfan, line 133: `bimUpload` is a bit misleading when you've factored this out to be a generic function

maybe just call it something like `uploadCmd`?

test/src/unit/repo_test_utils.h
```diff
@@ -20,6 +20,8 @@
 #include "repo_test_database_info.h"
 #include "repo_test_fileservice_info.h"
 #include <fstream>
+#include <repo/lib/repo_log.h>
```
carmenfan, line 23: no longer needed?

test/src/unit/repo_test_utils.h
```diff
@@ -20,6 +20,8 @@
...
+#include "../../bouncer/src/repo/error_codes.h"
```
carmenfan, line 24: no longer needed?

carmenfan (changes requested):
@har00n-haider looks fine in general, just some minor clean up related stuff!

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
har00n-haider: This line is required to match up a texture to its parent material nodes. It's used later down the line when a texture is parsed https://github.com/3drepo/3drepobouncer/blob/6f5db153802ccb66afd39eb354756a7eabcfbdcb/bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp#L188

carmenfan: @har00n-haider  I don't mean adding into `textureIdToParents`

I mean line 144, the return just before the end of the function

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
har00n-haider: nice catch, updated the the data type. The size to be wary of would be 4b though no (uint32 max)? ( kinda like how NTFS systems can only address a file under 4 GB right?)

carmenfan: true, was thinking int32_t!

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
har00n-haider: good call, this is done.

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
har00n-haider: oops, I dunno why my eyes were drawn to that other line....

I'll remove the return!

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
har00n-haider: I have changed it to an if/else chain. Unfortunately C++ doesn't allow switching on a string (dangit...). Ideally we could write an enum for the various JSON properties, but as you suggested at the start of this ticket, I tried to keep the refactoring fairly light.

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
har00n-haider: it has been erased!

test/src/system/st_3drepobouncerClient.cpp
har00n-haider: its done

test/src/unit/repo_test_utils.h
har00n-haider: indeed, its removed.

test/src/unit/repo_test_utils.h
har00n-haider: its removed

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
har00n-haider: I have updated the method to do this here:
https://github.com/3drepo/3drepobouncer/blob/b18deadb51eec7122568223e98e775d5a145a670/bouncer/src/repo/core/model/bson/repo_bson_factory.cpp#L939

this problem still exists for material nodes that still use the clone method but the fix for this would require geometry to be parsed before materials, I made a ticket for this [here ](https://github.com/3drepo/3drepobouncer/issues/416)

har00n-haider:
@carmenfan , feedback from the review has been implemented ready for another look!

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
```diff
@@ -893,37 +893,34 @@ TextureNode RepoBSONFactory::makeTextureNode(
 	const uint32_t    &byteCount,
 	const uint32_t    &width,
 	const uint32_t    &height,
-	const int         &apiLevel)
+	const int         &apiLevel,
```
carmenfan, line 896: Can we swap the order? parentIDs go before apiLevel, every other function in this file that accepts a parentID array take it this way.

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
```diff
@@ -893,37 +893,34 @@ TextureNode RepoBSONFactory::makeTextureNode(
...
+	const std::vector<repo::lib::RepoUUID>* parentIDs)
```
carmenfan, line 897: This doesn't need to be a pointer? make it a ref instead would be better i think (or at least inline with everything else)
```suggestion
	const std::vector<repo::lib::RepoUUID> &parentIDs)
```

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
```diff
@@ -937,6 +934,21 @@ TextureNode RepoBSONFactory::makeTextureNode(
 		repoWarning << " Creating a texture node with no texture!";
 	}
 
+	//--------------------------------------------------------------------------
```
carmenfan, line 937: appendDefaults will sort this out for you if you pass the parentIDs in, you don't have to create your own function to put it into the document:
image

carmenfan (changes requested):

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
har00n-haider: this broke code on the assimp loader , but i refactored it to match

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
har00n-haider: done.

Just a side question though:
is it worth removing dupes and ordering before passing it in? This is what the original call to cloneAndAddParent was doing:
https://github.com/3drepo/3drepobouncer/blob/37f7ac3ad8cf13bd403e323ce956378573d87c35/bouncer/src/repo/core/model/bson/repo_node.cpp#L87

har00n-haider:
@carmenfan changes are completed

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
carmenfan: probably not. cloneAndAddParent is doing this because it has to merge the new parents with the existing parents, therefore it need to ensure there's no duplicates.

In this case in repoBsonFactory we're instantiating a new instance from scratch so we don't really have this issue

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
carmenfan: Fair enough, I'm just going to tweak that quickly, the apiLevel stuff are kind of useless legacy that we don't really use. even if we do it should just be using the default, so I've just altered to function call so it doesn't pass it in

carmenfan (approved):

PR #407, opened 2020-11-23
Fixes Issue 406

sebjf:
Compresses the binary buffer section of the SRC file using zlib with Boost stream filters. Indicates this to the application by adjusting the SRC header magic value (as this is no longer a standards compliant SRC).

### DIFF
--- a/bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_src.cpp
+++ b/bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_src.cpp
@@ -23,6 +23,10 @@
 #include "../../../core/model/bson/repo_bson_factory.h"
 #include "../../../lib/repo_log.h"
 #include "../../modelutility/repo_mesh_map_reorganiser.h"
+#include <iostream>
+#include <boost/iostreams/filtering_streambuf.hpp>
+#include <boost/iostreams/copy.hpp>
+#include <boost/iostreams/filter/zlib.hpp>
 
 using namespace repo::manipulator::modelconvertor;
 
@@ -103,6 +107,8 @@ SRCModelExport::SRCModelExport(
 	const repo::core::model::RepoScene *scene
 	) : WebModelExport(scene)
 {
+	enableCompression = true;
+
 	//Considering all newly imported models should have a stash graph, we only need to support stash graph?
 	if (convertSuccess)
 	{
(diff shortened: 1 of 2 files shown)


bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_src.cpp
```diff
@@ -103,6 +107,8 @@ SRCModelExport::SRCModelExport(
 	const repo::core::model::RepoScene *scene
 	) : WebModelExport(scene)
 {
+	enableCompression = true;
```
carmenfan, line 110: use REPO_BOOST_NO_GZIP, like in `repo_model_import_3drepo.cpp`

I think for our current use case we'd always want to gzip this. the only situation is if zlib isn't compiled into boost (default windows boost doesn't come with it, I had to manually compile it  :( ).

the REPO_BOOST_NO_GZIP flag is used so we can still compile 3drepobouncer if it doesn't exist, which is probably more appropriate than a hard coded flag

bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_src.cpp
```diff
@@ -164,8 +176,25 @@ std::unordered_map<std::string, std::vector<uint8_t>> SRCModelExport::getSRCFile
...
+				boost::iostreams::filtering_streambuf<boost::iostreams::input> out;
+				out.push(boost::iostreams::zlib_compressor());
+				out.push(boost::iostreams::array_source((const char*)fullDataArray.data(), fullDataArray.size()));
+
+				// The first 4 bytes define the length of the uncompressed data. ZLib, at the least, requires this be known ahead of time.
+				std::vector<uint8_t> compressed;
+				compressed.resize(4);
+				((uint32_t*)compressed.data())[0] = fullDataArray.size();
+				
+				compressed.insert(compressed.end(), std::istreambuf_iterator<char>(&out), std::istreambuf_iterator<char>());
```
carmenfan, line 191: would it not make more sense to just insert the size number into the fullDataArray, at compressed.begin(), instead of copying the whole of the data array into compressed?

carmenfan (changes requested):

bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_src.cpp
sebjf: Hi @carmenfan , my understanding is that the [filtering_streambuf](https://www.boost.org/doc/libs/1_73_0/libs/iostreams/doc/classes/filtering_streambuf.html) will perform the compression during the iteration (insert into compressed); the array_source makes fullDataArray appear as a [device](https://www.boost.org/doc/libs/1_74_0/libs/iostreams/doc/classes/array.html). As the "device" is unbuffered, I am not sure what would happen if we tried to write the compressed data back into fullDataArray.
(If I understood what you meant correctly of course!)

bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_src.cpp
sebjf: @carmenfan that does make more sense. Fixed.

bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_src.cpp
carmenfan: @sebjf Ah ok, I didn't realised i'ts doing a compression whilst inserting. It just looked like we're inserting a massive buffer into a tiny buffer instead of the other way round!

carmenfan (approved):

PR #430, opened 2021-02-02
Issue 428 unhandled promise fixes + unity queue improvement

### LINKED ISSUE #428: unhandled promise rejection in message decoder
### Description
Messages can get stuck in the queue if for some reason bouncer_worker cannot find the instruction json.

image

We need to catch it and return error message

### Goals
- [x] Should fail at init if shared_dir doesn't eixst
- [x] if *.json doesn't exist, it should catch the error gracefully and return an error message back on callbackq
- [x] on exitAfter mode, the process should wait for the predefined time limit for a new task before exiting

### Tasks
- [x] At init, check if shared_dir exists as part of the health test
- [x] catch promise rejection from message decoder, send apprioriate error message back to callbackq

Also extra requests:
- [x] On exit after N task mode, introduce a timer to wait for tasks before exiting.

carmenfan:
This fixes #428
#### Description
- tasks should no longer get stuck in the queue due to instruction file missing
- shared_dir and log_dir are checked at init, process will exit if they do not exist
- Changed unity editor call to depict the full name of unity.log and run_unity.log
- add logic to wait for a task in _exitAfter_ mode if the queue is empty
- introduced 2 new config variables:
    - `pollingIntervalMS` In _exitAfter_ mode, the number of ms to wait inbetween polling the queue for tasks (default: 10000)                                                                                                                                                     
    - `maxWaitTimeMS` In _exitAfter_ mode, the maximum wait time for a task before exiting the process (default: 300000 - aka 5mins)

### DIFF
--- a/tools/bouncer_worker/src/lib/utils.js
+++ b/tools/bouncer_worker/src/lib/utils.js
@@ -15,6 +15,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>
  */
 
+const moment = require('moment');
+
 const Utils = {};
 
 Utils.exitApplication = (errCode = -1) => {
@@ -25,5 +27,7 @@ Utils.exitApplication = (errCode = -1) => {
 Utils.sleep = (ms) => new Promise((resolve) => {
 	setTimeout(resolve, ms);
 });
+
+Utils.getCurrentDateTimeAsString = () => moment().format('DD-MM-YY_HH[h]mm[m]ss[s]');
 
 module.exports = Utils;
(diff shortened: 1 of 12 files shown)


tools/bouncer_worker/src/lib/utils.js
```diff
@@ -26,4 +28,6 @@ Utils.sleep = (ms) => new Promise((resolve) => {
 	setTimeout(resolve, ms);
 });
 
+Utils.getCurrentDateTimeAsString = () => moment().format('DD-MM-YY_HH[h]mm[m]ss[s]');
```
keifgwinn, line 31: Can we format this YYYY-MM-DD_HH[h]mm[m]ss[s] please to be an iso recognised date format.

keifgwinn (approved):
just the datetime stamp is confusing

Issue #432, opened 2021-02-08
Texture Coordinates not considered by GeometryCollector when matching vertices


sebjf:
In the GeometryCollector::addFace method, the GeometryCollector will generate face indices by matching the face vertices to existing ones based on their content. This method takes uv coordinates as arguments, but only considers position and normal attributes when matching. This creates the possibility that two or more vertices with different UVs but the same position & normals degenerate to one of them, and the uv coordinates of the others are lost.

(In practice this is very unlikely, because two vertices with the same position and normals, but different uvs, would mean a UV seam on a planar surface, which would be a very odd choice when unwrapping UVs, either automatically or manually.)

sebjf:
When this addressed we should also look at the GeometryCollector::meshData map. Currently when a mesh format changes a new mesh is added (rather than finding one with an existing format). If ODA, say, interleaved triangles and lines, this could result in an unnecessarily large number of nodes.

sebjf:
> When this is addressed we should also look at the GeometryCollector::meshData map. Currently when a mesh format changes a new mesh is added (rather than finding an existing one with a matching format). If ODA, say, interleaved triangles and lines, this could result in an unnecessarily large number of nodes.

This has been resolved already in #429.

PR #438, opened 2021-02-25
Issue 429

### LINKED ISSUE #429: Add Support for 2D Lines - Milestone 3
Add support for importing 2D lines from DGN files via ODA.

sebjf:
This fixes #429. The DataProcessor superclass has been modified to override the polyline callbacks from ODA, in order to receive 2D lines. The GeometryCollector has also been modified to support multiple mesh formats.


bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
```diff
@@ -79,94 +81,203 @@ repo::core::model::RepoNodeSet repo::manipulator::modelconvertor::odaHelper::Geo
...
-		//Start a new mesh if the current mesh has UVs but this face doesn't or vice versa.
-		auto lastMeshName = nextMeshName;
-		stopMeshEntry();
-		nextMeshName = lastMeshName + "_";
-		repoDebug << "Staring a new mesh due to UV difference";
-		startMeshEntry();
+	if (!vertices.size())
+	{
+		repoError << "Vertices size[" << vertices.size() << "] is unsupported. A face must have more than 0 vertices";
+		exit(-1);
```
carmenfan, line 236: ideally if we can do some error handling that gets propagate all the way back up that'll be great.

Reason being the bouncer_worker takes the error code from the program and it's base on https://github.com/3drepo/3drepobouncer/blob/master/bouncer/src/repo/error_codes.h

Using those will give us a better idea of why something failed (which ultimately propagates all the way back up to the web UI to show the user)

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
```diff
@@ -79,94 +81,203 @@ repo::core::model::RepoNodeSet repo::manipulator::modelconvertor::odaHelper::Geo
...
 	}
 
-	if (faceHasUV && uvCoords.size() != vertices.size()) {
+	auto meshData = startOrContinueMeshByFormat(getMeshFormat(uvCoords.size(), true, vertices.size()));
+
+	if ((uvCoords.size() > 0) && uvCoords.size() != vertices.size()) {
 		repoError << "Vertices size[" << vertices.size() << "] and UV size [" << uvCoords.size() << "] mismatched!";
 		exit(-1);
```
carmenfan, line 243: Here too

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
```diff
@@ -79,94 +81,203 @@ repo::core::model::RepoNodeSet repo::manipulator::modelconvertor::odaHelper::Geo
...
 	entry.layerName = nextLayer.empty() ? "UnknownLayer" : nextLayer;
-
+	entry.format = format;
 	return entry;
 }
 
-void  GeometryCollector::startMeshEntry() {
-	nextMeshName = nextMeshName.empty() ? std::to_string(std::time(0)) : nextMeshName;
+void GeometryCollector::startMeshEntry() {
+	nextMeshName = nextMeshName.empty() ? boost::lexical_cast<std::string>(uuidGenerator()) : nextMeshName;
```
carmenfan, line 95: we have a wrapper around boost uuid, might be better to use that instead? 

then you can just do 
```suggestion
	nextMeshName = nextMeshName.empty() ? RepoUUID::createUUID().toString() : nextMeshName;
```

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
```diff
@@ -79,94 +81,203 @@ repo::core::model::RepoNodeSet repo::manipulator::modelconvertor::odaHelper::Geo
...
 
-	if (faceHasUV && uvCoords.size() != vertices.size()) {
+	auto meshData = startOrContinueMeshByFormat(getMeshFormat(uvCoords.size(), true, vertices.size()));
+
+	if ((uvCoords.size() > 0) && uvCoords.size() != vertices.size()) {
 		repoError << "Vertices size[" << vertices.size() << "] and UV size [" << uvCoords.size() << "] mismatched!";
 		exit(-1);
 	}
 
+	// Todo: this method attempts to generate indices by matching vertex based on attributes, but doesn't yet match the uv coordinates.
```
carmenfan, line 246: This doesn't break revit textured models right?

Would we be in a situation of joining up vertices with different UVs?

carmenfan (changes requested):
code doesn't compile in linux, can you take a look? https://travis-ci.com/github/3drepo/3drepobouncer/jobs/487445433
image

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
sebjf: @carmenfan Yes! This the bug we spoke about the other week. I've raised a maintenance issue here: https://github.com/3drepo/3drepobouncer/issues/432. The reason we won't have hit it yet is because to trigger it, there would have to be collocated vertices, with identical normals, but different UVs - that is, a UV map seam on a flush surface.
This would be a very very odd thing to do, because in that case one would just adjust the UV coordinates at the edges themselves to move, say, a decal, around.  
(Even in the case where one had something like a screen or poster on a wall it wouldn't be done, because only a single texture is applied per UV set, so changing the content would involve changing the UVs. Far easier to just to change the texture.)

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
sebjf: @carmenfan OK, so shall I update this exit call to return an appropriate error code, or throw an exception that gets caught further up (or similar)? 
(I was wondering about this as ODA (I am fairly sure it comes from ODA) throws a "exited with fatal code" exception during teardown when the debugger is attached, even though the exit code is 0!)

carmenfan: Whatever works and is easiest is fine!

sebjf: Seems like no matter what we do ODA intercepts our exit code or exception, and re-issues it with its own that obscures ours. This might be because the DataProcessor is a "device" to ODA, so ODA is designed not to allow it to throw or otherwise try and take down the process.
In any case, I've added an error code member to the geometry collector that is checked by OdaModelImport.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
sebjf: Resolved above!

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
```diff
@@ -79,94 +81,203 @@ repo::core::model::RepoNodeSet repo::manipulator::modelconvertor::odaHelper::Geo
...
+
+	return currentMesh;
+}
+
+void GeometryCollector::addFace(
+	const std::vector<repo::lib::RepoVector3D64>& vertices)
+{
+	if (!vertices.size())
+	{
+		repoError << "Vertices size[" << vertices.size() << "] is unsupported. A face must have more than 0 vertices";
```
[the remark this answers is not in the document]
sebjf: Updated!

carmenfan (approved):

PR #444, opened 2021-03-08
Issue 443 - Pure 2D Lines Model Import Failure

### LINKED ISSUE #443: Pure 2D files are coming out with import failure if the user has src Enabled
### Description
As we don't support polylines right now, SRC exporter is throwing out an error saying no geometry in the file, resulting in an import failure.

image

### Steps to replicate
- Ask andrew for the model (or just find a model with only polylines in it)
- upload the file

### Current Behaviour
- application exits with error code 32

### Expected Behaviour
- skip SRC generation if there are no 3D meshes, the application should continue to import so we can generate unity bundles

sebjf:
This fixes #443 

`SceneManager::commitScene` decides whether to generate SRCs based on `shouldGenerateSrcFiles`. This method checks for meshes, but does not discriminate primitive type. This PR updates that check to make it explicit.

Note that due to the way the srcAssets.json is generated, models with lines will show errors on loading into Unreal (404s for the missing line supermesh srcs). It is safe to ignore these until they can be fixed properly (by updating, io, bouncer or the plugin).

### DIFF
--- a/bouncer/src/repo/manipulator/modelutility/repo_scene_manager.cpp
+++ b/bouncer/src/repo/manipulator/modelutility/repo_scene_manager.cpp
@@ -515,9 +515,29 @@ bool SceneManager::shouldGenerateSrcFiles(
 	const repo::core::model::RepoScene* scene,
 	repo::core::handler::AbstractDatabaseHandler* handler) const
 {
-	repo::core::model::RepoUser user(handler->findOneByCriteria(REPO_ADMIN, REPO_SYSTEM_USERS, BSON("user" << scene->getDatabaseName())));
 	//only generate SRC files if the user has the src flag enabled and there are meshes
-	return scene->getAllMeshes(repo::core::model::RepoScene::GraphType::DEFAULT).size() && user.isSrcEnabled();
+	
+	bool hasTriangleMeshes = false;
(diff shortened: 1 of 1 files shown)


bouncer/src/repo/manipulator/modelutility/repo_scene_manager.cpp
```diff
@@ -515,9 +515,29 @@ bool SceneManager::shouldGenerateSrcFiles(
...
+
+		if (mesh->getPrimitive() == repo::core::model::MeshNode::Primitive::TRIANGLES)
+		{
+			hasTriangleMeshes = true;
+			break;
+		}
+	}
+
+	repo::core::model::RepoUser user(handler->findOneByCriteria(REPO_ADMIN, REPO_SYSTEM_USERS, BSON("user" << scene->getDatabaseName())));
+	bool userHasSrcFlag = user.isSrcEnabled();
```
carmenfan, line 538: @sebjf can we check for this first?

If the user doesn't have the SRC flag then we can return early and would not have to probe the meshes.

carmenfan (changes requested):

bouncer/src/repo/manipulator/modelutility/repo_scene_manager.cpp
sebjf: @carmenfan Updated!

carmenfan (approved):

PR #449, opened 2021-03-17
ISSUE #432 - Vertex Referencing in GeometryCollector

### LINKED ISSUE #432: Texture Coordinates not considered by GeometryCollector when matching vertices

sebjf:
This fixes #432. 

When adding faces, the GeometryCollector class attempts to reference existing vertices if possible, before adding new ones, but the previous implementation only considered vertices and an optional normal. This meant texture coordinates may be mismatched, and there was a whole second pathway for lines.

This PR adds a new VertexMap class that performs vertex buffer construction & referencing, updates the GeometryCollector class to use this, and do to so through one addFace overload.

### DIFF
--- a/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
+++ b/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
@@ -156,118 +156,65 @@ mesh_data_t* GeometryCollector::startOrContinueMeshByFormat(uint32_t format)
 void GeometryCollector::addFace(
 	const std::vector<repo::lib::RepoVector3D64>& vertices)
 {
-	if (!vertices.size())
-	{
-		repoError << "Vertices size[" << vertices.size() << "] is unsupported. A face must have more than 0 vertices.";
-		errorCode = REPOERR_ODA_GEOMETRY_ERROR;
-		return;
-	}
-
-	auto meshData = startOrContinueMeshByFormat(getMeshFormat(false, false, vertices.size()));
-
-	repo_face_t face;
-	for (auto i = 0; i < vertices.size(); ++i) 
-	{
-		auto& v = vertices[i];
-		int vertIdx = 0;
-
-		// match the vertex to generate an index for this face
-		if (meshData->vToVIndex.find(v) == meshData->vToVIndex.end())
-		{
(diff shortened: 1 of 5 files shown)


bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
```diff
@@ -155,119 +155,66 @@ mesh_data_t* GeometryCollector::startOrContinueMeshByFormat(uint32_t format)
 
 void GeometryCollector::addFace(
 	const std::vector<repo::lib::RepoVector3D64>& vertices)
+{
```
carmenfan, line 158: do we still need this function? could we  just apply default parameters on the function it's calling with false?

carmenfan (commented):
looks fine in general, just a question below

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
sebjf: We could - I deliberately left it as an overload because not all combinations of the most general `addFace` are valid (e.g., uvs without normals). So the public methods correspond to valid uses, and the general one is private (with the intent that if someone tries to use it in another way in the future, they'll find its private and be forced to look at the implementation and so the comment that says there are limitations on how it's to be used!)

If you think the overload obfuscates it more than the alternative though, I can change it to use default parameters?

carmenfan: @sebjf  I think it's fine as is, just looked a bit weird at first glance that it has to send false false to the other function.

But I think your argument of the separation and the latter function being a private function makes sense!

carmenfan (approved):

PR #455, opened 2021-03-24
#Issue_453: 3d lines support for BIM

### LINKED ISSUE #453: lines support for BIM
Linked to https://github.com/3drepo/3drepoPlugin/issues/231

- [x] version update
- [x] check for type of mesh when creating node
- [x] test with file from repoFileExporter visual unit test 
- [x] Fix issue with the error on import (missing node)

har00n-haider:
This fixes #453 

## Description
- Adding support for Line primitive types in the mesh nodes
- Parsing lineweight and adding it to the material node
- Using error code for unsupported primitive types
- Slight improvement in the way file versions are handled
## Related tickets
- https://github.com/3drepo/3drepoPlugin/issues/243
- [Revit 2d lines from 3d view BIM export](https://github.com/3drepo/3drepoPlugin/issues/231)
- [AssetBundleCreator bug fix](https://github.com/3drepo/3drepounity/issues/366)

## Testing
 - [x] Add any new test files to the tests repo
### Manual 
- [x] If an object have 2D lines and 3D faces it should still show up as a single object in 3drepo.io
### Unit
- [x] BIM v1 file should error out 
- [x] BIM v2 file processes with no errors
- [x] BIM v3 file processes with no errors
- [x] BIM v4 file with 2D lines should import successfully
- [x] BIM v4 file of 3D geometry should import successfully
- [x] Serialisation and deserialition test of the updated repomaterialnode_t 
- [x] Double checked any references to size of the repomaterialnode_t as its changed
### System
- [x] ST for BIMv4 to make sure the overall pipeline is all good


bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -285,15 +287,43 @@ RepoModelImport::mesh_data_t RepoModelImport::createMeshRecord(
...
+			else
+			{
+				primitiveType = MeshNode::Primitive::TRIANGLES;
+			}
 
-				faces.push_back(tmpFace);
+			// pull out faces
+			switch (primitiveType)
+			{
+			case repo::core::model::MeshNode::Primitive::LINES:
```
carmenfan, line 305: I think it's better to just merge the 2 cases and use variables, as they're essentially doing the same thing hence quite a bit of duplicated code.

ie something like this (untested, unchecked):

```suggestion
			case repo::core::model::MeshNode::Primitive::LINES:
			case repo::core::model::MeshNode::Primitive::TRIANGLES:
			        const auto length = (int8_t)primitiveType;
				for (int i = 0; i < numIndices; i += length)
				{
					repo_face_t tmpFace;
					for (int j = i; j < i + length; ++j)
						tmpFace.push_back(tmpIndices[j]);
					faces.push_back(tmpFace);
				}
				break;
```

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -285,15 +287,43 @@ RepoModelImport::mesh_data_t RepoModelImport::createMeshRecord(
...
+				{
+					repo_face_t tmpFace;
+					tmpFace.push_back(tmpIndices[i]);
+					tmpFace.push_back(tmpIndices[i + 1]);
+					tmpFace.push_back(tmpIndices[i + 2]);
+					faces.push_back(tmpFace);
+				}
+				break;
+			default:
+				//TODO: how to error out here?
```
carmenfan, line 325: @sebjf  added a new error code for identifying import errors regarding unexpected geometry content:  REPOERR_ODA_GEOMETRY_ERROR

Should probably rename to REPOERR_GEOMETRY_ERROR and use that.

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -432,7 +462,9 @@ bool RepoModelImport::importModel(std::string filePath, uint8_t &err)
 		repoInfo << "Loading BIM file [VERSION: " << incomingVersion << "]";
 
 		size_t metaSize = REPO_VERSION_LENGTH + sizeof(fileMeta);
-		if (incomingVersion == REPO_V3)
+		//TODO: Clean up the version string/number/comparison stuff into a class
```
carmenfan, line 465: I think either do it, or log a maintenance issue against it.

Inline todos just gets forgotten about and never get done 😆

carmenfan: though a class for version comparison seems a bit over kill?

carmenfan (changes requested):
it's not reading the line width?

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
carmenfan: also since you're no longer testing against the strings, you may as well remove the strings from the header and have 2 int depicting the min/max supported versions

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -432,7 +462,9 @@ bool RepoModelImport::importModel(std::string filePath, uint8_t &err)
...
+		uint8_t incomingVersionNo = std::stoi(incomingVersion.substr(3,3));
+		if (incomingVersionNo > 2)
 		{
 			fin->read((char*)&file_meta, REPO_V3_FILEMETA_BYTE_LEN);
```
carmenfan, line 469: maybe just rewrite this to a static dictionary?

as `REPO_V3_FILEMETA_BYTE_LEN` is also the same as V3's byte length, the naming can be confusing to people.

i.e. 
```suggestion
// Have this array declared somewhere (maybe next to the min version supported variable)
const std::unordered_map<int, int> FILE_META_BYTE_LEN_BY_VERSION = {{2, 56}, {3, 72}, {4, 72}};

// Then you no longer need an if and this line becomes:
			fin->read((char*)&file_meta, FILE_META_BYTE_LEN_BY_VERSION [incomingVersionNo]);
```

You could even integrate this with the supported version check - if there's no entry in the dictionary for that number then it's an unsupported version.

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -26,7 +26,9 @@
 #include "../../../core/model/bson/repo_bson_factory.h"
 #include "../../../lib/repo_log.h"
 #include "../../../error_codes.h"
+#include "../../../core/model/bson/repo_node_mesh.h"
```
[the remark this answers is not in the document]
har00n-haider: I always went of the principle of having all the includes you need for a file explicitly set. Can remove it if you want.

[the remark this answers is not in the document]
carmenfan: if we do that then we would be adding material transformation etc etc

it just makes the file unnecessary long and it doesn't really do anything. It's also something very difficult to enforce as a lot of libraries (like mongo) gives you a centralised header to include and it links to everything else. under that principle you'd have to dig around every class you utilise and include that class header

carmenfan:
@har00n-haider  fyi I just merged staging into this branch, coverall is complaining that we dropped code coverage on the testing, but it looks like it could be because I increased it on staging due to the hotfix and this branch doesn't have that change yet.

har00n-haider:
@carmen, this PR is ready. One thing to note is I haven't written the unit test for the new BIM004 format (lines support), was going to do that once the revit plugin was stable. Let me know if you want me to add one it at this stage.

carmenfan:
@har00n-haider you may as well add it now, even if we can the section views we're keeping the 2D lines in 3D so this PR will so go in even if the revit one needs changing

carmenfan:
@har00n-haider just nit picking on the PR template 😆 


image

It's not about what you've tested with, it's more about expected outcomes (and any edge cases), i.e. something like this:
- A BIM v4 file of 3D geometry should import successfully
- A BIM v4 file with 2D lines should import successfully
- If an object have 2D lines and 3D faces it should still show up as a single object in 3drepo.io
- BIM v3 file should still work
- BIM v2 file should still work

carmenfan (changes requested):
We're still not reading the line width and saving it all the way into the the database. 

It should be reading the value from material struct and writing it into the bson object here: https://github.com/3drepo/3drepobouncer/blob/master/bouncer/src/repo/core/model/bson/repo_bson_factory.cpp#L123

And also the reverse process of reading it from the bson and putting it into the material struct here: https://github.com/3drepo/3drepobouncer/blob/master/bouncer/src/repo/core/model/bson/repo_node_material.cpp#L40

carmenfan:
Travis failed because you changed the error code name without changing it on the other places that are used

image

har00n-haider:
> Travis failed because you changed the error code name without changing it on the other places that are used
> 
> image
> 
> EDIT: I just made a commit to fix this so I can move on with testing the plugin!

Sorry about that I was having some trouble with the search function of VS back, it wasn't properly searching the whole solution. I have fixed the issue now tho.

bouncer/src/repo/lib/datastructure/repo_structs.h
```diff
@@ -104,6 +104,7 @@ typedef struct{
 	float opacity = 1;
 	float shininess = 0;
 	float shininessStrength = 0;
+	float lineWeight = 0;
```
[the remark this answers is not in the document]
carmenfan: I think we either need to ensure default lineWeight is one, or it defaults to NaN (so it's not written)

Otherwise other processes which doesn't read line weight (like the ODA importer) will stop seeing lines when we take this property into account!

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.cpp
```diff
@@ -126,14 +127,20 @@ void RepoModelImport::parseMaterial(const boost::property_tree::ptree &matTree)
 		textureId = matTree.get<int>("texture");
 	}
 
+	if (matTree.find("lineWeight") != matTree.not_found())
+	{
+		repo_material.lineWeight = matTree.get<float>("lineWeight");
+	}
```
carmenfan, line 133: we should set this to NaN if the property is not found?

carmenfan (changes requested):
@har00n-haider  just a couple of things regarding the line weight stuff

carmenfan (approved):

PR #461, opened 2021-04-12
ISSUE #460 update mongodb (nodejs) driver

### LINKED ISSUE #460: Update io driver on bouncer worker
### Description
Upgrade the mongodb nodejs driver we're using on bouncer_worker to the latest available

carmenfan:
This fixes #460
#### Description
- version bump mongo driver
- use `replaceOne` instead of `updateOne` as updateOne no longer let you replace the whole document

#### Test cases
- toy model (only functionality that uses the mongo driver) should still work

### DIFF
--- a/tools/bouncer_worker/package.json
+++ b/tools/bouncer_worker/package.json
@@ -15,7 +15,7 @@
   "dependencies": {
     "amqplib": "0.5.3",
     "moment": "2.29.1",
-    "mongodb": "2.2.24",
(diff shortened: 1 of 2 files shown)


carmenfan:
@Charence fyi when you look into upgrading mongo driver on 3drepo.io:

> use replaceOne instead of updateOne as updateOne no longer let you replace the whole document

update seems to require you use `$set` `$unset` operators now

PR #466, opened 2021-05-06
ISSUE_400 - Process monitoring to Elastic

### LINKED ISSUE #400: Option for detailed stats monitoring of processed files
### Description
We would like to have some more statistical information about when models are processed, and what they used.

total number of models processed
Type of model file 
File extension vs 
size vs 
Processing time
Storage requirement
Upload failures vs file type
Upload failures vs error code
Upload failures vs Account

### Goals
- [x] Design elastic record schema
- [x] Make configuration optional to emit to elastic

### Tasks

- [x] Establish processing monitoring capabilities
- [x] Gather all the information required into an elastic packet

keifgwinn:
This fixes #400

#### Description
- Adding new option to monitor processtime and memory utilisation and emit the record to an elasticsearch system

#### Test cases
- All normal bouncer functionality maintained
- Can create elastic records based on activity for 3drepobouncer or 3drepounity processes.


tools/bouncer_worker/src/queues/modelQueueHandler.js
```diff
@@ -54,7 +56,30 @@ Handler.onMessageReceived = async (cmd, rid, callback) => {
...
 
 	try {
-		returnMessage.value = await runBouncerCommand(logDir, cmdParams);
+		const memoryReporting = { enabled: config.elastic.memoryReporting, maxMemory: 0, processTime: 0 };
+		returnMessage.value = await runBouncerCommand(logDir, cmdParams, memoryReporting);
+		if (memoryReporting.maxMemory > 0) {
+			const importFile = cmdParams[3];
+			// eslint-disable-next-line
+			const fileStats = fs.statSync(require(importFile).file);
+			const elasticBody = {
```
carmenfan, line 65: as both unity and model queue need to generate this, i think it's best to have it in utils as a function.

So you just pass in the parameters, and it returns you the object.

This way, if we ever rename any of these, or if the body needs to be structured any differently, there's only 1 place to change.

tools/bouncer_worker/src/queues/modelQueueHandler.js
```diff
@@ -54,7 +56,30 @@ Handler.onMessageReceived = async (cmd, rid, callback) => {
 	};
 
 	try {
-		returnMessage.value = await runBouncerCommand(logDir, cmdParams);
+		const memoryReporting = { enabled: config.elastic.memoryReporting, maxMemory: 0, processTime: 0 };
+		returnMessage.value = await runBouncerCommand(logDir, cmdParams, memoryReporting);
+		if (memoryReporting.maxMemory > 0) {
+			const importFile = cmdParams[3];
```
carmenfan, line 62: instead of reading the config file again, you should just get decoder to return it to you. 

just add  `file: file: cmdFile.file` in here: https://github.com/3drepo/3drepobouncer/blob/master/tools/bouncer_worker/src/lib/messageDecoder.js#L59

tools/bouncer_worker/src/queues/unityQueueHandler.js
```diff
@@ -21,9 +21,11 @@ const { generateAssetBundles, validateUnityConfigurations } = require('../tasks/
 const { ERRCODE_ARG_FILE_FAIL, ERRCODE_BUNDLE_GEN_FAIL } = require('../constants/errorCodes');
 const { UNITY_PROCESSING } = require('../constants/statuses');
 const logger = require('../lib/logger');
+const Elastic = require('../lib/elastic');
 
 const logLabel = { label: 'UNITYQ' };
 
+// eslint-disable-next-line max-len
```
carmenfan, line 28: do you need this? since you didn't change the next line and it never complained about it before?

tools/bouncer_worker/src/queues/modelQueueHandler.js
```diff
@@ -54,7 +56,30 @@ Handler.onMessageReceived = async (cmd, rid, callback) => {
...
+				Teamspace: user,
+				Model: model,
+				Database: database,
+				MaxMemory: memoryReporting.maxMemory,
+				ProcessTime: memoryReporting.processTime,
+				DateTime: Date.now(),
+				// eslint-disable-next-line
+				FileType: require(importFile).file.split('.').pop().toString(),
+				FileSize: fileStats.size,
+				Process: '3drepobouncer',
```
carmenfan, line 75: avoid hard coded strings.

maybe have this in utils and reference it?
````js
Utils.PROCESS_TYPE = {
    bouncerClient: "3drepobouncer",
//etc
}

````

File: tools/bouncer_worker/src/queues/modelQueueHandler.js [hunk 5]
```diff
@@ -54,7 +56,30 @@ Handler.onMessageReceived = async (cmd, rid, callback) => {
...
 	try {
-		returnMessage.value = await runBouncerCommand(logDir, cmdParams);
+		const memoryReporting = { enabled: config.elastic.memoryReporting, maxMemory: 0, processTime: 0 };
+		returnMessage.value = await runBouncerCommand(logDir, cmdParams, memoryReporting);
+		if (memoryReporting.maxMemory > 0) {
+			const importFile = cmdParams[3];
+			// eslint-disable-next-line
+			const fileStats = fs.statSync(require(importFile).file);
+			const elasticBody = {
+				Teamspace: user,
```
carmenfan, line 66: teamspace should be the same as database.

user is the person who uploaded the file. database is where it's going to.

e.g. if `carmen` uploads a file in `keif`'s teamspace. user = `carmen`, teamspace & database = `keif`

tools/bouncer_worker/src/queues/unityQueueHandler.js
```diff
@@ -34,7 +36,26 @@ const processUnity = async (database, model, user, logDir, modelImportErrCode) =
 
 	try {
 		if (database && model) {
-			await generateAssetBundles(database, model, logDir);
+			const memoryReporting = { enabled: config.elastic.memoryReporting, maxMemory: 0, processTime: 0 };
+			await generateAssetBundles(database, model, logDir, memoryReporting);
+			if (memoryReporting.enabled) {
+				const elasticBody = {
+					Teamspace: user,
```
carmenfan, line 43: should be database here

tools/bouncer_worker/src/lib/elastic.js
```diff
@@ -0,0 +1,168 @@
...
+		auth: {
+			username: ELASTIC_CLOUD_AUTH[0],
+			password: ELASTIC_CLOUD_AUTH[1],
+		},
+		reload_connections: true,
+		maxRetries: 5,
+		request_timeout: 60,
+	};
+	const elasticClient = new Client(config);
+	elasticClient.cluster.health({}, (err, resp) => {
```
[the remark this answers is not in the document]
carmenfan: Also, we should probably do a health check on elastic connection at launch?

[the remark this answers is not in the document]
carmenfan: another thought... is it better to just create a client and keep using it, instead of creating a new client everytime this function is called?

tools/bouncer_worker/src/lib/elastic.js
```diff
@@ -0,0 +1,168 @@
...
+		if (err) logger.debug(`[ELASTIC]:  health ${resp.toString()}`, Elastic.loglabel);
+	});
+	return elasticClient;
+};
+
+Elastic.createElasticRecord = async (elasticClient, index, elasticBody, id, mapping) => {
+	try {
+		// eslint-disable-next-line no-param-reassign
+		id = id || Utils.hashCode(Object.values(elasticBody || {}).toString());
```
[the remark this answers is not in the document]
keifgwinn: The function was developed originally as a general purpose elastic handler, so sometimes we want to be able to set the ID but most of the time we don't care and want it generated for us.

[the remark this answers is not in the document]
carmenfan: if you want to keep it then I suggest making a variable for it so we avoid disabling the linter.

tools/bouncer_worker/src/lib/elastic.js
```diff
@@ -38,131 +52,61 @@ Elastic.createElasticClient = () => {
...
+
+const createElasticRecord = async (elasticIndex, elasticBody, id, mapping) => {
 	try {
-		// eslint-disable-next-line no-param-reassign
-		id = id || Utils.hashCode(Object.values(elasticBody || {}).toString());
-		const indexName = index.toLowerCase(); // requirement of elastic that indexs be lowercase
-		const configured = await elasticClient.indices.exists({ index: indexName });
-		if (!configured.body) {
+		const elasticClient = await elasticClientPromise;
+		let internalID = id;
```
carmenfan, line 71:
```suggestion
		const internalID = id || Utils.hashCode(Object.values(elasticBody || {}).toString());
```

tools/bouncer_worker/src/lib/processMonitor.js
```diff
@@ -1,59 +1,139 @@
...
 	try {
-		const data = Number(fs.readFileSync('/sys/fs/cgroup/memory/memory.usage_in_bytes'));
-		if (pidArray.length > 0) {
-			stats = await pidusage(pidArray);
-			if (data > maxMemory) maxMemory = data;
+		data = await si.osInfo();
+	} catch (e) {
+		logger.error('Failed to get operating system information record', logLabel);
+	}
+	return data.platform;
```
carmenfan, line 25: if catch happens, this will break the program as you'd be accessing `platform` of undefined.

tools/bouncer_worker/src/lib/processMonitor.js
```diff
@@ -1,59 +1,139 @@
...
+};
+const currentOSPromise = getCurrentOperatingSystem();
+
+const getDockerEnvironment = async () => {
+	const currentOS = await currentOSPromise;
+	if (currentOS === 'linux' && fs.existsSync('/.dockerenv')) {
```
carmenfan, line 31:
```suggestion
	return currentOS === 'linux' && fs.existsSync('/.dockerenv');
```

tools/bouncer_worker/src/lib/processMonitor.js
```diff
@@ -1,59 +1,139 @@
...
+};
+const isDockerEnvironment = getDockerEnvironment();
+
+const getCurrentMemUsage = async () => {
+	let data;
+	let result;
+	try {
+		if (isDockerEnvironment) {
+			// required to get more accurate information in docker
+			result = Number(fs.readFileSync('/sys/fs/cgroup/memory/memory.usage_in_bytes'));
```
carmenfan, line 44: this will get overwritten by L47 anyway?

tools/bouncer_worker/src/lib/processMonitor.js
```diff
@@ -1,59 +1,139 @@
...
 		}
-		} catch (err) { logger.verbose(`[processMonitor][maxmem][error]: ${err}`, processMonitor.logLabel); }
+		data = await si.mem();
```
carmenfan, line 46:
```suggestion
		result = (await si.mem()).used;
```

tools/bouncer_worker/src/lib/processMonitor.js
```diff
@@ -1,59 +1,139 @@
...
+	const currentOS = await currentOSPromise;
+	if (currentOS === 'linux' && fs.existsSync('/.dockerenv')) {
+		return true;
+	}
+	return false;
+};
+const isDockerEnvironment = getDockerEnvironment();
+
+const getCurrentMemUsage = async () => {
+	let data;
```
carmenfan, line 39: data is not used outside of the try {}, infact it's never overwritten so it should just be a `const`

But with the suggested change you don't even need a variable.

tools/bouncer_worker/src/lib/processMonitor.js
```diff
@@ -1,59 +1,139 @@
...
+	}
+	return result;
+};
+
+const maxmem = async () => {
+	let data;
+	let pids;
+
+	try {
+		data = await getCurrentMemUsage();
```
carmenfan, line 60: as far as I can tell you don't overwrite data again, it's also not used outside of this scope. so it should just be a const here.
```suggestion
		const data = await getCurrentMemUsage();
```

tools/bouncer_worker/src/lib/processMonitor.js
```diff
@@ -1,59 +1,139 @@
...
+
+		// add the missing properties for sending.
+		informationDict[stopPID].ReturnCode = returnCode;
+		informationDict[stopPID].MaxMemory = workingDict[stopPID].maxMemory - workingDict[stopPID].startMemory;
+		logger.verbose(`${stopPID} ${workingDict[stopPID].maxMemory} - ${workingDict[stopPID].startMemory} = ${informationDict[stopPID].MaxMemory}`, logLabel);
+		informationDict[stopPID].ProcessTime = workingDict[stopPID].elapsedTime;
+
+		if (elasticEnabled) {
+			try {
+				Elastic.createRecord(informationDict[stopPID]);
```
carmenfan, line 124: you need await here?

tools/bouncer_worker/src/lib/runCommand.js
```diff
@@ -4,40 +4,34 @@ const { ERRCODE_TIMEOUT, ERRCODE_UNKNOWN_ERROR } = require('../constants/errorCo
...
 	{ codesAsSuccess = [], verbose = true, logLabel },
-	memoryReporting = { enabled: false, maxMemory: 0, processTime: 0 },
+	processInformation,
 ) => new Promise((resolve, reject) => {
-	if (verbose) logger.info(`Executing command: ${exe} ${params.join(' ')} memoryReporting: ${memoryReporting.enabled}`, logLabel);
+	if (verbose) logger.info(`Executing command: ${exe} ${params.join(' ')} processMonitoring: ${processReporting}`, logLabel);
 	const cmdExec = spawn(exe, params, { shell: true });
-	if (memoryReporting.enabled) {
-		processMonitor.startMonitor(cmdExec.pid);
+	if (processReporting) {
```
carmenfan, line 17:
```suggestion
	if (processReporting && !processInformation.doNotMonitor) {
```

carmenfan: instead of having a special flag for do not monitor, isn't it better to just pass undefined for processInformation to denote that?

carmenfan (changes requested):

keifgwinn:
We should add the versions of software doing the processing to the logs so that we can track performance

carmenfan:
@keifgwinn  I've made quite a bit of change on this. In theory it should work pretty much like before, but I can't 100% guarantee as i only tested process monitoring on windows and i didn't test elastic. So you probably want to run this before we merge it in.

Just a note for the future, a few of my previous comments were marked resolved when they weren't resolved - can we make sure we don't do that, otherwise that means I can't trust things are done and have to go through everything from scratch again. It's ok if you don't agree with the change, but make a comment and don't mark it as resolve so I know there's a conversation going 😆 

### Config changes
I made this more align with the changes on 3drepo.io:
- Elastic config is now expected to be exactly what you pass into elastic api. no elastic object = don't initialise elastic.
- logger now have `noColors` `jsonOutput` options

### Elastic code
This has been made to be pretty much identical to the 3drepo.io version

### Process monitoring
I've rewrote the monitoring as it was excessively complex and doesn't really do what I think you want it to do. 
- I removed `pidusage` dependency and all the extra logic with tracking if the process is still running. From what I can tell all they do is to see whether we should keep going with the timer - which is a bit pointless since we know when to start/stop them as we call start/stop monitor immediately before/after.
- I rewrote timeout to setInterval (like the comment I made the first time round)... Before it was launching a timeout everytime we have a process to monitor, and every time the timer is triggered it go through all the pids we want to track. meaning if there's 100 parallel process - we would've launched 100 timers, every one of them would've tried to track data from 100 pids. Not that it'd ever happen, and also the validity of the memory data also goes down the drain if we have more than 1 process running. I've changed it so it now creates an interval timer for every startMonitor call, which only track deal with the process it called from (well, kind of - since memory calculation is not pid related it's not really). In theory it should be no different to what you were doing, but it should be cleaner than what you had before.

carmenfan (dismissed):
Can merge, pending final testing

carmenfan (approved):

keifgwinn:
tested and working nicely (and possibly better than before)

carmenfan (approved):

PR #483, opened 2021-07-20
ISSUE #452 detect unity license errors

### LINKED ISSUE #452: Unity is faking success on model creation when there's a licensing issue
### Description
During model import, if UnityEditor didn't execute the function due to missing licenses, it's currently returning a retval 0.

Meaning bouncer worker thinks it's a success. and then tells 3drepo.io so.

So the User would think the model uploaded successfully, but sees 0 revision in the list.

We need to figure a way to track this behaviour and return a special error code, and act accordingly

carmenfan:
This fixes #452
#### Description
- detect unity license error with a very ad hoc approach (grepping the log)
- if detected, it will override the current retVal and return with a new error code (35).
- This err code is captured on the unity queue handler, if it is true, then it will reject the promise of `onMessageReceived`, which the queue handler will return to the queue with a `nack` - hence it will be put back in the queue for retry.

- This error will be printed on the log as follows: `Failed to generate asset bundle: Invalid unity license`
image

#### Test cases
- normal processes should work as before
- if the box has unity licensing error, it should no longer pretend to succeed and will requeue the message.

### DIFF
--- a/bouncer/src/repo/error_codes.h
+++ b/bouncer/src/repo/error_codes.h
@@ -85,3 +85,5 @@
 #define REPOERR_TOY_IMPORT_FAILED 33
 //The scene was read successfully but failed during geometry processing due to content problem
 #define REPOERR_GEOMETRY_ERROR 34
+//Used in bouncer worker only
+#define REPOERR_UNITY_LICENSE_INVALID 35
--- a/tools/bouncer_worker/src/constants/errorCodes.js
+++ b/tools/bouncer_worker/src/constants/errorCodes.js
@@ -24,6 +24,6 @@ module.exports = {
 	ERRCODE_ARG_FILE_FAIL: 16,
 	ERRCODE_TIMEOUT: 29,
 	ERRCODE_TOY_IMPORT_FAILED: 33,
-	BOUNCER_SOFT_FAILS: [7, 10, 15], // failures that should go through to generate bundle
-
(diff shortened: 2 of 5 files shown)


keifgwinn:
Looks like it's working for me on a linux build, however it would be handy if when it encountered this it would sleep for 2 minutes or so. At the moment it hits retry very fast, so sends a lot of updates to the IO side with 'generating bundles' and is also running the unity command every second or so, which makes it hard to license the machine in situ. If we had a bit of a pause, then I've got time to get in and run the licensing command 'live' and then copy out the results for persisting.

keifgwinn:
It also managed to break the build on one because it was retrying so fast that unity didn't even log a proper full response that included the bad license message, and so it then broke in the traditional success/fail unity

keifgwinn (changes requested):
Technically works fine, but behaviour when we fall into this state isn't great.

A delay of 5 minutes when it enters this state would be good, so that there's time to actually license the unity before it retries again, and would not spam back at the IO side constantly that there's bundles being generated.

Also, once or twice it failed the job out because the unity logs were not complete, which I think was part of the speedy retrying process.

carmenfan:
@keifgwinn  I added a sleep of `rabbitmq.maxWaitTimeMS` (which is used for the max wait time for next job before exit, and defaults to 5mins), does that work? 
image

PR #486, opened 2021-08-03
ISSUE #485 Absorb the bouncer wrapper code and add option to build alongside bouncer

### LINKED ISSUE #485: Move bouncer wrapper compilation to 3drepobouncer
### Description
At the moment the bouncerwrapper is a separate compilation inside 3drepounity repository. This purely serves as an interop layer between C++ to external applications that are not written in C++.

It is historically placed there because assetBundleCreator is the only thing that uses it. However, the dependency is purely on the 3drepobouncer side, as it does not require any code from the unity repository.  This occasionally causes problems where 3drepobouncer and 3drepounity may go out of sync, and the developer has not recompiled 3drepounity, thus the unity processing crashes as the 3drepobouncer dll no longer has the same interface that 3drepobouncerWrapper was built with.

also with the addition of AssetBundleCreatorHelper, the process of setting up the assetBundleCreator now takes so many steps and is becoming more cumbersome. 

It would be good if we move the bouncer wrapper source code to be part of 3drepobouncer, and is part of the build process (as one of the options), so we can be certain that everytime you compile 3drepobouncer, bouncer wrapper is also updated and the bindings are always correct.

linked to https://github.com/3drepo/3drepounity/issues/395

### Goals
- Move bouncerwrapper sourcecode
- add additional flag in 3drepobouncer compilation to determine whether bouncer wrapper should be compiled.
- if flagged, bouncerwrapper should be compiled and placed in th dll folder (similar process as before)

### Tasks
- [x] place the wrapper code base under ./wrapper (everything under 3drepounity\BouncerWrapper)
- [x] update the bouncer Cmakelists to add a target for the wrapper 
- [x] update the wrapper Cmakelists  to remove elements no longer required (i.e. library detection etc.)
- [x] add a flag in the bouncer Cmakelists to optiononally add a project for the wrapper dll
- [x] place the output in the BOUCNCER_INSTALL\lib directory
- [x] update the bouncerwrapper function to pull the version directly from the dll
- [x] ammend the cmake call in the travis yaml to build the bouncerwrapper dll
- [x] add a line in the travis to run a some test on the wrapper function to confirm the critical functions used by the unity ABC work as expected
- [x] confirm that travis can succesfully build and test the bouncerwrapper dll
- [x] make the wrapper tests [dependant](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html) on whether the wrapper was built
- [x] delete readme / shift to wiki

har00n-haider:
This fixes #485 

linked to https://github.com/3drepo/3drepounity/pull/396

## Description

Bouncerwrapper exposes the bouncer API in a way that is easily consumable from managed code (e.g. C#). Having the wrapper code build as part of the core bouncer reduces the chances of the wrapper API being out of sync with core bouncer API.

- It has been moved to the this repository
- The CMake build process has been updated to provide an option to build it 
- The CMake build process also provides an option to build a gtest suite for the wrapper (dependant on whether the wrapper is being built). No additions were made to the "tests" repo.
- The CMake build process will only build the gtest framework if either of the test target options are checked
- Slight editing of the wrapper code to 
  - Provide the same version as the core bouncer
  - Move function definitions from the wrapper header to a cpp file (due to dllimport giving an error if applied to a function definition)
- Updating the wikis: [bouncerwiki](https://github.com/3drepo/3drepobouncer/wiki/3drepobouncer-explained), [mainwiki](https://github.com/3drepo/wiki/wiki/3drepo.io-Setting-up-a-local-instance-for-development#building-bouncerwrapper).

## Remaining tasks
- [x] Confirm that travis can succesfully build and test the bouncer wrapper

## Test cases
- Unit tests pass to confirm all the wrapper API functions used in the critical path of the asset bundle creator
- The asset bundle creator can run with the newly built wrapper directly from the bouncer repo (used in conjunction with local instance of 3drepo.io)
- The bouncer can built without bouncerwrapper + bouncerwrappertest Cmake options
- The original bouncer test can be built without the bouncerwrapper + bouncerwrappertest Cmake options


har00n-haider:
@carmenfan travis is failing on the 3drepobouncerwrapperTest due to a segfault. Good news is I guess it found the executable!

On my local machine the tests pass with a fresh restore of the db from tests. Could be that the 3drepobouncerTest is making changes to the db that the 3drepobouncerwrapperTest is not expecting? I am running that scenario on my local machine atm.

Do you think its worth me testing this locally in linux VM? Or maybe just rerunning on travis with some more debug messages in the test itself...

carmenfan:
> @carmenfan travis is failing on the 3drepobouncerwrapperTest due to a segfault. Good news is I guess it found the executable!
> 
> On my local machine the tests pass with a fresh restore of the db from tests. Could be that the 3drepobouncerTest is making changes to the db that the 3drepobouncerwrapperTest is not expecting? I am running that scenario on my local machine atm.
> 
> Do you think its worth me testing this locally in linux VM? Or maybe just rerunning on travis with some more debug messages in the test itself...


it's probably worth sorting out the failures before investigating on the seg fault. Could the seg fault be that you make assumptions of previous function call gives you valid data so it becomes corrupted when you make that assumption as you carry on? if that's the case you might want to use ASSERT_TRUE or ASSERT_EQ instead of EXPECT so the test just ends if any of the assertions were not met

carmenfan (changes requested):

carmenfan (approved):

carmenfan (approved):

carmenfan:
@keifgwinn  about to merge this in. just FYI after this is merged the bouncerwrapper is going to be compiled as part of 3drepobouncer (along with Client etc)

so there shouldn't be a need to run the CMAKE stuff on the unity build (it will still need the built binaries from the bouncer compilation to work)

PR #498, opened 2021-08-25
Issue 496 - fix killing process after timeout

carmenfan:
This fixes ##496

#### Description
`spawn.kill()` seems to have stopped working on v14. using `tree-kill` instead.


keifgwinn (approved):
LGTM, tested behaviour on staging and everything's working as expected.

PR #520, opened 2021-11-29
Issue#519 adding a 3drepo license check stage

### LINKED ISSUE #519: Add license check
https://github.com/3drepo/3drepounity/issues/414

har00n-haider:
This fixes #519

# Description
- Updated the bouncer worker to take in a new error code and provide it to rabbit queue
- Added the error code to the bouncer error code list
- Added CMAKE variables to pass the required license information to the compilation stage
- Updated CMAKE with a find module for the cryptolense libs
- Crypto root directory is assumed to be:
  - **Root**
    - **include**: crypto headers
    - **lib**: built binaries
      - **Debug**
      - **Release**
    - **third_party**: source code dependencies of crypto, including header files that are required
- class added to deal with the activation/deactivation of the license , called from the constructo/descructor of the repocontroller respectively. See class comments for details
- Removed VALID_UNTIL options from the CMAKE, replaced with CRYPTOLENS_LICENSE instead. All other license configurations (expiry dates etc. are done in the cryptolense portal)
- Removed the version info function in repo controller as we do not currently cache the LicenseKey info, and there is no C++ way of a `GetKey` style function that only returns LicenseKey info (as opposed the activation)
- Moved the logging setup to the repocontroller constructor  as the license activation code needs it
- Moved all implementation to the repo controller impl class
- Updated the bouncer wrapper to destroy the repocontroller on parent process exit (explicitly call the deactivation function)

# Remaining work
- [x] Compile the libs from cryptolens
- [x] Add the dependency to that lib in the bouncer CMAKE (with the provision to disable the licensing)
- [x] Create the licences class in the same way that the c# code is made
- [x] Call the activation from where the killswitch is currently checked
- [x] Update the CMAKE to include a platform switch to choose which crypto libraries to link against
- [x] Make sure the license deactivation works in the same way as in the [.NET implementation](https://github.com/3drepo/3drepounity/pull/415)
- [x] Test the version checked bouncer build on linux (see references below for more info)
- [x] have Cmake throw an error when the licenes check is enabled but the any of license related variables are unset 
- [x] We need to throw an error if the third_party folder is not found within CRYPTOLENS_ROOT
- [x] When the license activation failed due to expired license, make sure its not occupying a seat.
- [x] Create a solution that replicates the issue seen when the bouncer dll does not deregister
  - [x] confirm bug occurs in bouncer test 
  - [x] confirm bug occurs in abc
  - [x] confirm bug does not occur when running bouncer client directly
- [x] convert the repoInfo to traces
- [x] Update internal documentation on the new requirements to build bouncer [Here](https://github.com/3drepo/wiki/wiki/3drepo.io-Setting-up-a-local-instance-for-development#licensing)
- [x] Call the deactivation where the program exits
- [x] Remove the killswitch once the license check is verified
- [x] Confirm that expired codes do not work 
- [x] Confirm with the full pipeline 3drepo.io pipeline
- [x] Move definitions to cpp
- [x] Remove the crypto depencies from the test project, and make sure that crypto checks are disabled 
- [x] bouncertools still compiles correctly
- [x] auto generate a UUID in bouncer worker and pass through abc/bouncer via env vars (chek for the env var, if it does not exist then auto generate an instance locally)

# Test cases
 
Tests that passed are marked

- [x] License not present: FAIL
- [x] Instance limit reached:  FAIL
- [x] License not expired, is blocked - FAIL
- [x] License is expired, is blocked - FAIL
- [x] License is expired, not blocked - FAIL
- [x] License not expired, not blocked - PASS
- [x] All license information is logged the bouncer INFO stream
- [x] All failures are propagated up to .io
- [x] bouncer client deactivates the token succesfully on exit
- [ ] wrapper should deactivate the license when used via the unity project,
> Currently an exception is being thrown by the cryptolens library in some scenarios (bouncerWrapper has an error on deactivation,  wheras bouncerClient is fine). There is a continuing line of communication with cryptolense regarding this.
- [x] Using the same license to activate/deactivate consecutivley in ABC and in bouncerWrapper occupying only one seat

# Notes

- Assimp include directory is not being set correctly for me despite me setting the root env var. Assimp lib directory is succesfully found though. 
- Some bouncer upload tests are failing in windows. I will wait to see if the tests fail in linux aswell, before investigating further.

# Resources
- https://cryptolens.io/2017/08/new-client-api-for-c/
- https://github.com/Cryptolens/cryptolens-cpp


har00n-haider:
@carmenfan a question about the bouncer test project, should update the Cmake to make sure that we test so that you can't build them  make it so that the tests never use crypto

bouncer/CMakeLists.txt
```diff
@@ -5,13 +5,21 @@ add_subdirectory(src)
...
 add_library(3drepobouncer SHARED ${SOURCES})
 set_target_properties(3drepobouncer PROPERTIES RELEASE_POSTFIX "_${VERSION_MAJOR}_${VERSION_MINOR}")
 set_target_properties(3drepobouncer PROPERTIES DEBUG_POSTFIX "_${VERSION_MAJOR}_${VERSION_MINOR}_d")
-target_link_libraries(3drepobouncer ${Boost_LIBRARIES} ${MONGO_LIBRARIES} ${ASSIMP_LIBRARIES} ${OCCT_LIBRARIES} ${IFCOPENSHELL_PARSERLIB} ${IFCOPENSHELL_GEOMLIB} ${ODA_LIB} ${SYNCHRO_READER_LIBRARIES} ${THRIFT_LIBRARIES} ${ZLIB_LIBRARIES} ${SYNCHRO_LIBRARIES} ${AWSSDK_LIBRARIES} ifcUtils_2x3 ifcUtils_4)
+
+# checks for platform specific libs
+set(LIBS ${Boost_LIBRARIES} ${MONGO_LIBRARIES} ${ASSIMP_LIBRARIES} ${OCCT_LIBRARIES} ${IFCOPENSHELL_PARSERLIB} ${IFCOPENSHELL_GEOMLIB} ${ODA_LIB} ${SYNCHRO_READER_LIBRARIES} ${THRIFT_LIBRARIES} ${ZLIB_LIBRARIES} ${SYNCHRO_LIBRARIES} ${AWSSDK_LIBRARIES} ${CRYPTOLENS_LIBRARIES} ifcUtils_2x3 ifcUtils_4)
+if(WIN32)
+# required for cryptolens on windows
+	Set(LIBS ${LIBS} Winhttp.lib)
```
carmenfan, line 19: add that to `CRYPTOLENS_LIBRARIES`

bouncer/src/repo/lib/repo_license.cpp
```diff
@@ -0,0 +1,169 @@
...
+*/
+
+#include "repo_license.h"
+
+namespace Licensing
+{
+	// definitions for class static variables
+	repo::lib::RepoUUID LicenseValidator::instanceId;
+	std::string LicenseValidator::licenseStr;
+	std::unique_ptr<Cryptolens> LicenseValidator::cryptolensHandle;
```
carmenfan, line 28: Can we make sure that the library still compiles without license check? I'm pretty sure it'll fail at the moment as Cryptolens wouldn't be a part of the included library.

bouncer/src/repo/lib/repo_license.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+	void LicenseValidator::RunActivation()
+	{
+#ifdef REPO_LICENSE_CHECK
+		// only run once 
+		if (!instanceId.isDefaultValue() && cryptolensHandle) return;
```
carmenfan, line 36: just store instanceId as string and just do `empty()` check.

bouncer/src/repo/lib/repo_license.cpp
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	void LicenseValidator::RunDeactivation()
+	{
+#ifdef REPO_LICENSE_CHECK
+		// only deactivate if we have succesfully activated
+		if (instanceId.isDefaultValue() && cryptolensHandle) return;
+
+		cryptolens::Error e;
+		cryptolensHandle->deactivate(
```
carmenfan, line 120: do we need to store `instanceId`? it doesn't look like we use it here? I guess `cryptolensHandle` already contain all the necessary information?

carmenfan: Also I think we want to deconstruct `cryptolensHandle` here and reset the variable.

Whilst it's fair to say that the current use case, the program is very likely to exit immediately, but it's probably more future proof to do it with the assumption that the user could potentially trigger the process again, instead of leaving the class in a weird state.

bouncer/src/repo/lib/repo_license.cpp
```diff
@@ -0,0 +1,169 @@
...
+		{
+			// printing out the result
+			std::string notes = licenseKey->get_notes().has_value() ?
+				licenseKey->get_notes().value() : "";
+			int noUsedInsances = licenseKey->get_activated_machines().has_value() ?
+				licenseKey->get_activated_machines()->size() : -1;
+			int maxInstances = licenseKey->get_maxnoofmachines().has_value() ?
+				licenseKey->get_maxnoofmachines().value() : -1;
+			bool licenseBlocked = licenseKey->get_block();
+			bool licenseExpired = licenseKey->check().has_expired(time(0)) ? true : false;
```
carmenfan, line 87:
```suggestion
			bool licenseExpired = licenseKey->check().has_expired(time(0));
```

bouncer/src/repo/lib/repo_license.cpp
```diff
@@ -0,0 +1,169 @@
...
+			(
+				e, // Object used for reporting if an error occured
+				authToken, // Cryptolens Access Token
+				productId, // Product key associated with the 3drepo.io pipeline
+				licenseStr, // License Key
+				floatingTimeIntervalSec // The amount of time the user has to wait before an actived machine (or session in our case) is taken off the license.
+			);
+
+		// dealing with early bail out scenarios
+		repoInfo << "****License activation summary****";
```
carmenfan, line 61:
```suggestion
		repoInfo << "****License activation summary****";
```

If we're adding a header block like this then we should add an end of block comment at the end of the function before we return.

bouncer/src/repo/lib/repo_license.cpp
```diff
@@ -0,0 +1,169 @@
...
+		cryptolens::Error e;
+		cryptolensHandle->deactivate(
+			e,
+			authToken,
+			productId,
+			licenseStr,
+			true);
+
+		// dealing with the error in deactivation
+		repoInfo << "****License deactivation summary****";
```
carmenfan, line 128: same here

bouncer/src/repo/repo_controller_internal.cpp.inl
```diff
@@ -38,6 +39,30 @@ RepoController::_RepoControllerImpl::_RepoControllerImpl(
 	{
 		subscribeToLogger(listeners);
 	}
+
+	// set logging directory
+	char* logDir = getenv("REPO_LOG_DIR");
```
carmenfan, line 44: Can we refactor `getenv` logic into repo_utils and get everywhere to just call that function? so it takes in a envVar to look for and return ` std::string`

It's getting to the point where we are doing a lot of duplicated code 😆

bouncer/src/repo/repo_controller_internal.cpp.inl
```diff
@@ -38,6 +39,30 @@ RepoController::_RepoControllerImpl::_RepoControllerImpl(
...
+	std::string logPath;
+	logPath = logDir ? std::string(logDir) : "./log/";
```
carmenfan, line 46: I don't think it's necessary to wrap a char* with the constructor, it should auto do it (I'm 100% not sure, give it a go)
```suggestion
	logPath = logDir ? logDir : "./log/";
```

cmake_modules/FindCRYPTOLENS.cmake
```diff
@@ -0,0 +1,64 @@
...
+		NAMES cryptolens
+		PATHS
+		${CRYPTOLENS_ROOT}/lib/Debug
+	)
+	set(CRYPTOLENS_LIBRARIES
+		debug ${CRYPTOLENS_LIBRARIES_DEBUG}
+		optimized ${CRYPTOLENS_LIBRARIES_RELEASE}
+		)
+endif()
+
```
carmenfan, line 52: Can you copy over the auto-detection parts as well from one of the other scripts?

test/CMakeLists.txt
```diff
@@ -2,16 +2,14 @@ add_subdirectory(src)
...
+remove_definitions(-DREPO_LICENSE_CHECK)
 
 include_directories(${gtest_SOURCE_DIR}/include ${gtest_SOURCE_DIR} src ../bouncer/src ../client/src ../
-	${Boost_INCLUDE_DIR} ${MONGO_INCLUDE_DIR} ${ASSIMP_INCLUDE_DIR}  ${OCCT_INCLUDE_DIR} ${IFCOPENSHELL_INCLUDE_DIR}  ${ODA_INCLUDE_DIR} ${AWSSDK_INCLUDE_DIR}  ${SYNCHRO_READER_INCLUDE_DIR} )
+	${Boost_INCLUDE_DIR} ${MONGO_INCLUDE_DIR} ${ASSIMP_INCLUDE_DIR}  ${OCCT_INCLUDE_DIR} ${IFCOPENSHELL_INCLUDE_DIR}  ${ODA_INCLUDE_DIR} ${AWSSDK_INCLUDE_DIR}  ${SYNCHRO_READER_INCLUDE_DIR})
 add_executable(3drepobouncerTest ${TEST_SOURCES} ${SOURCES})
 target_link_libraries(3drepobouncerTest gtest
-	${Boost_LIBRARIES} ${MONGO_LIBRARIES} ${ASSIMP_LIBRARIES} ${OCCT_LIBRARIES} ${IFCOPENSHELL_GEOMLIB} ${IFCOPENSHELL_PARSERLIB}  ${ODA_LIB}  ${SYNCHRO_READER_LIBRARIES} ${THRIFT_LIBRARIES} ${ZLIB_LIBRARIES} ${SYNCHRO_LIBRARIES} ${AWSSDK_LIBRARIES} ifcUtils_2x3 ifcUtils_4)
+	${Boost_LIBRARIES} ${MONGO_LIBRARIES} ${ASSIMP_LIBRARIES} ${OCCT_LIBRARIES} ${IFCOPENSHELL_GEOMLIB} ${IFCOPENSHELL_PARSERLIB}  ${ODA_LIB}  ${SYNCHRO_READER_LIBRARIES} ${THRIFT_LIBRARIES} ${ZLIB_LIBRARIES} ${SYNCHRO_LIBRARIES} ${AWSSDK_LIBRARIES} ifcUtils_2x3 ifcUtils_4 Winhttp.lib)
```
carmenfan, line 12: This isn't adding in the license libs, we should just add it as part of the license lib (if windows)

carmenfan (changes requested):

test/CMakeLists.txt
carmenfan: you'd also need to check bouncerTools and bouncerWrapper are compiling correctly.

CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ endif()
 option (REPO_BUILD_COVERAGE "Building with coverage report" OFF)
 option (ODA_SUPPORT "If ODA library is present" OFF)
 option (SYNCHRO_SUPPORT "If Synchro library is present" OFF)
-option (VALID_UNTIL "Set timestamp for how long (sec since epoch) this library should be valid for (-1 for unlimited) " -1)
+option (CRYPTOLENS_LICENSE "Build with cryptolens license check" $ENV{ENABLE_LICENSE_CHECK})
```
carmenfan, line 36: don't do that, just let it be OFF like everything else, it doesn't need to be an env var.

carmenfan: Let's only reference CRYPTO if it's specifically about their library. LICENSE_CHECK should suffice.

bouncer/src/repo/lib/repo_license.cpp
har00n-haider: this does not compile as their bool operator definition is marked explicit, but adding a static cast to a bool makes sure the returned class from `has_expired` correctly. I've done it that way to make it look neater.

bouncer/src/repo/lib/repo_license.cpp
har00n-haider: I assume you mean that once `RunDeactivation` triggers, that we should clean up the `instanceId` and `cryptolensHandle`, so that another `RunActivation` call can be run succesfully in the same process? That's a good catch, will clean that up. I will carry across this change to the .NET implementation aswell.

har00n-haider:
@carmenfan, the general implementation is complete, barring the unchecked points in the description. It's ready for testing in linux, please update the checkboxes once your done.

I will update the remaining issue with the cryptolens crashing on deactivation, once I hear back from them.

CMakeLists.txt
```diff
@@ -33,7 +33,14 @@ endif()
 option (REPO_BUILD_COVERAGE "Building with coverage report" OFF)
 option (ODA_SUPPORT "If ODA library is present" OFF)
 option (SYNCHRO_SUPPORT "If Synchro library is present" OFF)
-option (VALID_UNTIL "Set timestamp for how long (sec since epoch) this library should be valid for (-1 for unlimited) " -1)
+
+option (LICENSE_CHECK  "Build with a license check, see for details: https://github.com/3drepo/wiki/wiki/Licensing-components-of-3drepo.io " OFF)
+option (LICENSE_RSA_PUB_KEY_MOD "Licensing public token modulus" "")
```
carmenfan, line 38: these should only show if `LICENSE_CHECK` is on. move it to L126

`option` also probably isn't the right syntax, as the documentation depicts it to be an on and off thing. https://cmake.org/cmake/help/latest/command/option.html

we should be using set https://cmake.org/cmake/help/latest/command/set.html

bouncer/src/repo/lib/repo_license.h
```diff
@@ -0,0 +1,96 @@
...
+	static const int productId = LICENSE_PRODUCT_ID;
+
+#endif
+
+
+	/**
+	* Class that performs license checks given an license key taken from an
+	* environment variable.
+	* 
+	* The class implements a floating license model:
```
carmenfan, line 59: I don't think this comment block serves much purpose, the function header is pretty obvious what this does. It's a pretty bog standard activation/deactivation process.

If we keep it, remove the link and fix the typo in L68.

bouncer/src/repo/repo_controller_internal.cpp.inl
```diff
@@ -38,6 +39,29 @@ RepoController::_RepoControllerImpl::_RepoControllerImpl(
...
 		subscribeToLogger(listeners);
 	}
+
+	// set logging directory
+	std::string logDir = repo::lib::getEnvString("REPO_LOG_DIR");
+	std::string logPath = logDir.empty() ? "./log/" : logDir;
+	this->logToFile(logPath);
+
+	// set logging level
+	char* debug = getenv("REPO_DEBUG");
```
carmenfan, line 49: should be using `getEnvString` here now you've created a function.

Please do a find around the codebase to find any `getnev` and replace them

bouncer/src/repo/lib/repo_utils.h
```diff
@@ -30,16 +30,30 @@ namespace repo {
...
 			std::transform(str.begin(), str.end(), str.begin(), [](unsigned char c) { return std::tolower(c); });
 		}
 
-		static void toUpper(std::string &str) {
+		static void toUpper(std::string &str) 
+		{
 			std::transform(str.begin(), str.end(), str.begin(), [](unsigned char c) { return std::toupper(c); });
 		}
+
+		static std::string getEnvString(char const * const & envVarName)
```
carmenfan, line 48: can we take in a `std::string`, and just do a `c_str` in the function.

bouncer/src/repo/lib/repo_utils.h
```diff
@@ -30,16 +30,30 @@ namespace repo {
...
+		{
+			std::string envVarStr;
+			char* envVarStsPtr = getenv(envVarName);
+			if (envVarStsPtr && strlen(envVarStsPtr) > 0)
```
carmenfan, line 52: The variable and the assignment feels a bit redundant.

```suggestion
			return (envVarStsPtr && strlen(envVarStsPtr) > 0) ? envVarStsPtr : "";
```

bouncer/src/repo/lib/repo_utils.h
```diff
@@ -30,16 +30,30 @@ namespace repo {
...
-		static void toUpper(std::string &str) {
+		static void toUpper(std::string &str) 
+		{
 			std::transform(str.begin(), str.end(), str.begin(), [](unsigned char c) { return std::toupper(c); });
 		}
+
+		static std::string getEnvString(char const * const & envVarName)
+		{
+			std::string envVarStr;
+			char* envVarStsPtr = getenv(envVarName);
```
carmenfan, line 51: `envVarStsPtr` is a bit of a weird naming, i can't actually figure out what the `Sts` part means.

given the context of the function, we could just call it `value`

test/CMakeLists.txt
```diff
@@ -2,16 +2,14 @@ add_subdirectory(src)
 
 enable_testing()
 add_definitions(-DREPO_API_LIBRARY)
-
+# override as we do not want to impose license checks on the test project
```
carmenfan, line 5: I don't think this is necessary. By default it's `off`.

The user would have to make a conscious decision to turn it on, in which case it should be tested within the test case.

tools/bouncer_worker/src/tasks/unityEditor.js
```diff
@@ -66,7 +69,12 @@ UnityHandler.generateAssetBundles = async (database, model, rid, logDir, process
 		unityLog,
 	];
 
+	if (config.repoLicense) {
+		process.env.REPO_LICENSE = config.repoLicense;
+	}
+
 	try {
+    process.env.REPO_INSTANCE_ID = crypto.randomBytes(16).toString("hex");
```
carmenfan, line 77: considering you only need the instance ID if license is enabled, shouldn't it be inside L72?

carmenfan: generating randomBytes with `crypto` is a bit excessive, just use [uuid](https://www.npmjs.com/package/uuid)

tools/bouncer_worker/src/tasks/unityEditor.js
```diff
@@ -75,9 +83,16 @@ UnityHandler.generateAssetBundles = async (database, model, rid, logDir, process
...
-	}
+		switch (err) {
+			case ERRCODE_UNITY_LICENCE_INVALID:
+			case ERRCODE_REPO_LICENCE_INVALID:
+				throw err;
+			default:
+				throw await checkLicenceError(unityLog) ? ERRCODE_UNITY_LICENCE_INVALID : ERRCODE_BUNDLE_GEN_FAIL;
+		}
+	} finally{
+    delete process.env.REPO_INSTANCE_ID
```
carmenfan, line 94: There isn't much reason to do this.

it'll get overwritten eveytime something is ran.

if we do this then technically we should be doing this for every env var we're setting.

wrapper_test/src/unit/ut_repo_csharp_interface.cpp
```diff
@@ -37,56 +37,110 @@ static std::string getDataPath(
...
-	ASSERT_TRUE(loadedScene);
-	// asset bundle save check
+	// main loop to access supermesh data
+	int numSuperMeshes = repoGetNumSuperMeshes();
+	ASSERT_EQ(numSuperMeshes, 1);
+	repoInfo << "looping over supermeshes";
+	for (int superMeshIdx = 0; superMeshIdx < numSuperMeshes; superMeshIdx++)
+	{
+		// get info regarding the current supermesh
+		long numSubSupermeshes{0};
```
carmenfan, line 71: any reason why we're using the curly brace syntax over just doing a = 0?

har00n-haider: The curly brace (list initializer) does not allow type narrowing (e.g. float to int) but in this case does not make a difference. I'll change it to the equals syntax.

test/CMakeLists.txt
har00n-haider: can do this, but I have not confirmed if the tests pass with the licensing enabled.

tools/bouncer_worker/src/tasks/unityEditor.js
har00n-haider: I thought it might be too overkill to add a new package just for this, but I'll go ahead and do it that way.

carmenfan:
> @carmenfan the requested changes have been made. I cannot test fully as I am waiting on cryptolens to extend my trial period. Should be ready for testing on linux. I have updated our internal bouncer build instructions with the required changes for building with licensing [here](https://github.com/3drepo/wiki/wiki/3drepo.io-Setting-up-a-local-instance-for-development#licensing)

@har00n-haider code doesn't seem to compile right now 
image

har00n-haider:
> > @carmenfan the requested changes have been made. I cannot test fully as I am waiting on cryptolens to extend my trial period. Should be ready for testing on linux. I have updated our internal bouncer build instructions with the required changes for building with licensing [here](https://github.com/3drepo/wiki/wiki/3drepo.io-Setting-up-a-local-instance-for-development#licensing)
> 
> @har00n-haider code doesn't seem to compile right now image

@carmenfan apologies, I didn't have the oda option enabled so I missed that error. I have pushed a commit to fix it.

carmenfan (changes requested):
@har00n-haider at the moment, cmake passes if i check licenseCheck, but left all the license related variables unset.

image


Can we put some code in there to check they are not empty, otherwise throw an error?

carmenfan:
We  need to throw an error if the third_party folder is not found within CRYPTOLENS_ROOT

carmenfan:
**When the license activation failed due to expired license, it's occupying a seat**

I had an expired key, and when I run it for the first time, it's telling me that I have 1 activated license, but activation failed
image


I can see on the console that it is then occupying the seat:
image

If I run it again, it gives me a different message as the capacity has been reached:
image


I think we should be be deactivating before exit if this is the scenario

bouncer/src/repo/lib/repo_license.cpp
```diff
@@ -0,0 +1,207 @@
...
+				license, // License Key
+				floatingTimeIntervalSec // The amount of time the user has to wait before an actived machine (or session in our case) is taken off the license.
+			);
+
+		// dealing with early bail out scenarios
+		repoInfo << activationSummaryBlock;
+		if (e)
+		{
+			cryptolens::ActivateError error = cryptolens::ActivateError::from_reason(e.get_reason());
+			repoInfo << "- server message: " << error.what();
```
carmenfan, line 109: can we change these to repoTrace?

And just have a simple one line pass/failed message in `repoInfo`. (Also do this for deactivation)

Whilst it's nice to know the details, we don't care on an operational basis and it's quite a bit of clutter in the log, the user can run in verbose mode to get more if required

carmenfan:
@har00n-haider tested in linux and it seems to be working.

I had to make a few changes to make it compile, if you can quickly recompile on windows just to confirm I didn't break anything that'd be good.

har00n-haider:
@carmenfan feedback addressed and the sample solution has been sent to cryoto for their input. see [this](https://github.com/har00n-haider/3drepobouncer/tree/cryptoissue) branch for details.

carmenfan (approved):
Looks fine, waiting on devops to test

carmenfan:
@keifgwinn I just merged latest of staging in here (just incase..)

which contains the synchro v3 upgrade from the timezone branch.

So if you need to rebuild bouncer here it'll need the changes as per https://github.com/3drepo/DevOps/issues/321

carmenfan (approved):
@keifgwinn checked this and is happy with it working

PR #525, opened 2021-12-03
Issue 524 - replica set connection support

### LINKED ISSUE #524: Connect to Mongo replica set
### Description
At the moment, the configuration designed in such a way that we only take in one host and one port. We need to modify this so we can take in a replica set connection

### Config change
The config file that gets passed to bouncerClient will now accept `connectionString` , or the old `dbhost` and `dbport` (the latter takes precedence, so ensure they are removed if you are trying to use connection string.

Connection string should be as defined by mongo, for e.g. you can connect to an individual instance as like this:
image

### Goals
- To be able to connect to a replica set

### Tasks
- [x] Add the ability to instantiate a db handler with a connection string
   - [x]  Ensure errors are caught correctly if the string is not a valid connString.
   - [x] Ensure both repl Set and single instances can connect successfully
- [x] Wire the function above all the way up to RepoController
- [x] Alter bouncer config and bouncerClient to parse and understand this change 
  - [x] Ensure the old config still works for minimal disruption
- [ ] Clean up
  - [ ] There's a BSONCredential way of connecting, which i believe is for the old RepoGUI - which we can remove
  - [ ] Tests

carmenfan:
This fixes #524

#### Description
- Add an option to read database info from a connection string

#### Test cases
- old configurations should still work as before
- replacing `dbHost` and `dbPort` with `connectionString` should also work

### DIFF
--- a/bouncer/src/repo/manipulator/repo_manipulator.cpp
+++ b/bouncer/src/repo/manipulator/repo_manipulator.cpp
@@ -71,6 +71,25 @@ bool RepoManipulator::connectAndAuthenticateWithAdmin(
 	return handler != 0;
 }
 
+bool RepoManipulator::connectAndAuthenticateWithAdmin(
+	std::string       &errMsg,
+	const std::string &connString,
+	const uint32_t    &maxConnections,
+	const std::string &username,
+	const std::string &password,
+	const bool        &pwDigested
+)
+{
+	//FIXME: we should have a database manager class that will instantiate new handlers/give existing handlers
+	repo::core::handler::AbstractDatabaseHandler *handler =
+		repo::core::handler::MongoDatabaseHandler::getHandler(
+			errMsg, connString, maxConnections,
+			repo::core::handler::MongoDatabaseHandler::getAdminDatabaseName(),
+			username, password, pwDigested);
+
+	return handler != 0;
+}
+
 repo::core::model::RepoBSON* RepoManipulator::createCredBSON(
 	const std::string &databaseAd,
 	const std::string &username,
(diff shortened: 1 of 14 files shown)


carmenfan:
@3drepo/devops Can you guys test this on a replica set please?

the details on how to connect with a connection string is explained in the [issue description](https://github.com/3drepo/3drepobouncer/issues/524#issue-1069481454)

It should theoretically work as all I'm doing here is pass the connection string to the driver. The only concern is that we're using such an old driver that it may not understand a"modern" mongo connection string... As their documentation suggest you don't even need to specify the replicaSet name... http://mongocxx.org/api/legacy-1.1.2/classmongo_1_1ConnectionString.html

don't worry about putting username/authentication database on this (unless you have to, as that's still a separate step to the connection right now)

 (I'm not done yet so please don't merge, still need to do some clean up)

carmenfan:
@3drepo/devops decided against clean up - a lot of this will naturally go away when we upgrade the driver.

Can merge if you guys are happy with it.

keifgwinn:
yup, looks good

carmenfan:
@Charence  can you review it?

bouncer/src/repo/manipulator/repo_manipulator.cpp
```diff
@@ -71,6 +71,25 @@ bool RepoManipulator::connectAndAuthenticateWithAdmin(
...
+bool RepoManipulator::connectAndAuthenticateWithAdmin(
+	std::string       &errMsg,
+	const std::string &connString,
+	const uint32_t    &maxConnections,
+	const std::string &username,
+	const std::string &password,
+	const bool        &pwDigested
+)
+{
+	//FIXME: we should have a database manager class that will instantiate new handlers/give existing handlers
```
Charence, line 83: I guess this will be tackled in a separate issue

Charence: Doesn't `getHandler` do this job?

carmenfan: this is about a possibility of having 2 database connection to 2 different places.

it's a bit N/A as that's only a concern with 3drepogui, which is deprecated for years

Charence (approved):

Issue #518, opened 2021-11-29
Support for ODA BimNv Navisworks Import


sebjf:
### Description
Adding the ODA BimNv library and importer class to support import of Navisworks NWD files.

### Goals
- [x] Bouncer `import` command recognises NWD files
- [x] 3D geometry is imported
- [x] 2D geometry is imported
- [x] Geometry is transformed/aligned correctly
- [x] All metadata extracted by the Navisworks plugin is created in the same form as the plugin
- [x] The hierarchy in the viewer matches that in Navisworks
- [ ] Grid system is imported as lines

### Tasks
- [x] Integrate BimNv binaries (update cmake, etc) and make sure BimNv works with existing ODA kernel (can link)
- [x] Create an importer class modelled on Revit, DWG, etc
- [x] Integrate class into bouncer file import command logic
- [x] Have the class iterate over the file and create some 3D geometry (either a flat list, or the first encountered etc) 
- [x] Add support for transforms
- [x] Add support for 2D lines
- [x] Make sure all materials are converted correctly
- [x] Refactor to match Navisworks hierarchy
- [x] Add metadata support
- [x] Test if metadata matches that of the Plugin
- [ ] Add support for grid system
- [x] Create tests

### Remarks
BimNv is installed by copying the contents of `BimNv_vc14_amd64dll_22.11.zip` into the `teigha_2022.11` directory, where `teigha_2022.11` is given as the `ODA_ROOT` environment variable.

### Questions
1. Do we support NWCs?
2. Do we want to open NWFs and report an error or let it be caught by `OdaModelImport`
(relevant files are file_processor.cpp line 54 & repo_model_import_oda.cpp line 13)

sebjf:
Hi @mosment3DRepo, in Navisworks, the Arch Link Model has textures, but in the [reference import](https://staging.dev.3drepo.io/viewer/mosment3ddev/1d128090-478f-11ec-93c6-e3f089695f8b) it doesn't. Should the NWD import support textures?

mosment3DRepo:
My reference at the moment would be to ignore textures for now, working on the basis that they could create very heavy models and they aren't that useful.

Since Navisworks models are likely to be several smaller models federated we should aim to keep them as light as possible.

sebjf:
(Going to put the discussion about differences in here so we have a record!)

**Instanced Geometry**

[...]

PR #530, opened 2021-12-14
Issue 523 - Upgrade ODA to 22.11

### LINKED ISSUE #523: Upgrade ODA to 22.11
### Description
Upgrade ODA libraries to 22.11. This is because we need the latest version of BimNv to get the metadata from Navisworks files.

### Goals
- [ ] Bouncer compiles with the 22.11 ODA binaries
- [ ] All tests pass
- [ ] Import functionality is unchanged

### Tasks
- [ ] Get the 22.11 binaries for existing products
- [ ] Rebuild bouncer solution with new binaries
- [ ] Fix API changes

sebjf:
This fixes #523 

#### Description
This PR updates the Revit Data Processor to be compatible with BimRv 22.11. The breaking changes in this version are concerned with how parameters, specifically units, are handled. Whereas before a set of compile time constants were used to describe units/parameter types, now they are described at runtime using instances of a single, dedicated type.

These changes update how the metadata is extracted to use this new way of dealing with parameters & units.

There are five main changes in import behaviour:
1. Parameters that were previously resolved to either handles or stringified enums are now exported using proper localised strings.
2. The labels of some parameters for older Revit files (<2020) have changed. The new behaviour is consistent with Revit 21/22.
3. Some parameter fields are missing compared to the previous version of BimNv. It appears these were always null values.
4. Invalid or unloaded `ImportInstance` geometry is no longer drawn. This matches the behaviour of Revit 21/22.
5. Textures and materials that were not previously supported now appear.
6. Some UV coordinates now take scaling into account.

#### Test cases
The upgraded version of ODA has been tested against the following .rvt files:
* Complex Geometry_Chapel Roof
* House with textures_2021
* rac_advanced_sample_project
* Revit Sample Project
* Sample House_MEP With Section Box
* Sample House_With Custom Parameters

A visual inspection between the models imported by the previous and upgraded bouncer was performed in .io. A programmatic  comparison of the metadata between the versions was also performed.

#### Other Resources
[Discussion](https://forum.opendesign.com/showthread.php?20493-Linked-file-no-longer-being-vectorized-in-22-1) with ODA regarding drawing Unloaded linked files.


bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_rvt.cpp
```diff
@@ -345,518 +370,760 @@ void DataProcessorRvt::initLabelUtils() {
...
 
-std::string DataProcessorRvt::unitsToString(const OdBm::DisplayUnitType::Enum &units) {
-	switch (units) {
-	case  OdBm::DisplayUnitType::Enum::DUT_UNDEFINED:
-	case  OdBm::DisplayUnitType::Enum::DUT_CUSTOM:
-	case  OdBm::DisplayUnitType::Enum::DUT_GENERAL:
-	case  OdBm::DisplayUnitType::Enum::DUT_FIXED:
-	case OdBm::DisplayUnitType::Enum::DUT_CURRENCY:
+boost::optional<std::string> DataProcessorRvt::unitsToString(const OdBmForgeTypeId& units) {
+	if (units.isEmpty()) { // Use if not if..else in this function because otherwise we reach compiler limits on nested blocks
```
carmenfan, line 374: I'm assuming they don't provide a units to units label of some sort still? 😢 

I tried looking for one before I coded this, and failed

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_rvt.cpp
```diff
@@ -936,15 +1213,15 @@ std::unordered_map<std::string, std::string> DataProcessorRvt::fillMetadata(OdBm
...
 	{
 		fillMetadataByElemPtr(element, metadata);
 	}
 	catch (OdError& er)
 	{
-		repoInfo << "Caught exception whilst trying to fetch metadata by element pointer " << convertToStdString(er.description());
-
-		repoTrace << "Caught exception whilst trying to fetch metadata by element pointer " << convertToStdString(er.description());
+		repoInfo << "Caught exception whilst trying to fetch metadata by element pointer " << convertToStdString(er.description()) << " Code: " << er.code();
+		repoTrace << "Caught exception whilst trying to fetch metadata by element pointer " << convertToStdString(er.description()) << " Code: " << er.code();
```
carmenfan, line 1224: I think we can remove this line... not sure why we print it twice on 2 different severity 😆

carmenfan (changes requested):
@sebjf Can we also add a rvt 2022 file into the test ? since we adding support for it

carmenfan:
@sebjf  travis is struggling to compile, cmake couldn't find a lot of the libraries that we were using

 it looks like a few libraries have been renamed/moved/deleted?

keifgwinn:
seb sent me another update to the library which I've updated the travis links with but it is still failing.

carmenfan:
> seb sent me another update to the library which I've updated the travis links with but it is still failing.

The latest failures aren't related to finding the library, it's code. 

@sebjf it looks like it's struggling to convert a const char* to the boost::optional type you defined:

image

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_rvt.cpp
sebjf: Done!

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_rvt.cpp
sebjf: > I tried looking for one before I coded this, and failed

I thought that had come from a sample or at least there was a reason we were overriding, not that it forced that to be written by hand! 😧 

> I'm assuming they don't provide a units to units label

There is now a way to do this using `getLabelForSymbol`, though its very obscure...

Passing the unit spec objects directly from the parameter descriptor fails silently, because while they describe the unit, they don't actually describe a _symbol_ itself (which is what that method is expecting) (nor can they return the symbol, which would be useful...)
It turns out the format options object for the unit type/spec object needs to be retrieved, *that* object will return another spec type object, but this one for the symbol itself, and *this* object can be passed to `getLabelForSymbol` to return a valid string.

This mechanism is what it is used in the `OdBmSampleLabelUtilsPE::format` method, which fortunately we have the source to, so I've implemented something similar for `processParameter`.

sebjf:
> @sebjf Can we also add a rvt 2022 file into the test ? since we adding support for it

Added!

carmenfan (approved):
Looks fine and seems to run fine.

Passing it onto @3drepo/devops  to deploy to get @3drepo/product-team  to test

carmenfan:
Testing & code merge will happen in #518  as #518 contains all changes from this branch.

PR #537, opened 2022-01-07
 Issue 526: Support for timezones when importing synchro sequences

### LINKED ISSUE #526: Offset Synchro Sequences with a specified timezone
### Description
This is related to https://github.com/3drepo/3drepo.io/issues/2892

### Goals
- Read in timezone and country information passed in from the model upload json.
- Offset all sequence times by the timezone indicated
- Offset all activity times by the timezone indicated
  - Ensure offset is also applied on dates within the metadata 

### Tasks
<To be filled in by the developer>

har00n-haider:
This fixes #526

# Description
- Added a utility class to convert a "seconds from unix epoch" timestamp in a given timezone to its utc equivalent
- Uses this lib: https://github.com/HowardHinnant/date

# Test cases
- See unit tests for the utility function

# Notes
- inputs/outputs std::chrono::seconds which can be created like this `std::chrono::seconds s{1641523367};` to ensure type safety at compile time.


har00n-haider:
@carmenfan ready for a look/hooking up with the rest of the spm processing flow.

carmenfan (changes requested):
The boost CSV has 383 entries
image

When the wiki has 595
image

Quick google also says boost's date_time library is not great, even boost locale don't use them 
image

The first trouble I ran into is that `America/Toronto` doesn't exist (at least in the version of boost we're using) Can we investigate on using `icu` instead (or something else that you can find that is more reliable)

har00n-haider:
Hey @carmenfan , these are the two options I mentioned earlier:

- https://github.com/unicode-org/icu
- https://github.com/HowardHinnant/date

carmenfan:
@har00n-haider 

if you want to try it then I say give it a go

Considering we need to compile the cpp for the tz library we should still compile it separately and link it as usual, it's always nicer to have a layer of abstraction to third party database afterall.

But before you do all that, it's probably good to do a sandbox test to make sure it does what we need  (i.e. the numbers are coming out right for Daylight saving etc) and it's not too fussy to build/link up.

har00n-haider:
@carmenfan ready for a look

carmenfan:
@3drepo/product-team  approved as per https://github.com/3drepo/3drepo.io/pull/2920#pullrequestreview-863378863

carmenfan (approved):

PR #544, opened 2022-01-13
ISSUE #543 fix connections for import toy

### LINKED ISSUE #543: toy project needs to connect using dbAddress if it is provided
### Description
when we made changes for #524, we didn't[ add the support on toy import.](https://github.com/3drepo/3drepobouncer/blob/master/tools/bouncer_worker/src/tasks/importToy.js#L257)

Thus if the deployment passes in a `connectionString` instead of `dbhost` and `dbport`, it fails to connect.

### Steps to replicate
- setup bouncer worker connection to connect with a dbAddress instead of dbhost and dbport
- trigger toy import
- 
### Current Behaviour
- fails with undefined:undefined in the mongo url

### Expected Behaviour
- should pass as before

carmenfan:
This fixes #543

#### Description
- automatically generate `connectionString` (if undefined) when we initialise the config file 
- connection to mongo for `importToy` features now connect using `connectionString`

#### Test cases
- import toy should work:
    - when mongo connection string is given in the config instead of host and port
    - when  host and port are defined instead of mongo connection string

### DIFF
--- a/tools/bouncer_worker/src/lib/config.js
+++ b/tools/bouncer_worker/src/lib/config.js
@@ -49,6 +49,11 @@ const applyDefaultValuesIfUndefined = (config) => {
 	// processMonitoring
 	config.processMonitoring = config.processMonitoring || {};
 	config.processMonitoring.memoryIntervalMS = config.processMonitoring.memoryIntervalMS || 100;
+
+	// create connection string for db
+	if (!config.db.connectionString) {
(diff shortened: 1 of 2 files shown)


keifgwinn:
tested and importing again nojw

Charence (approved):

keifgwinn (approved):

PR #546, opened 2022-01-18
Issue 435: gtest upgrade

### LINKED ISSUE #435: GTest namespace deprecation (Upgrade GTest)
The gtest projects fail to build as these projects treat warnings as errors. The warning in this case is

`Warning	C4996	'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED. You can define _SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING to acknowledge that you have received this warning.`

The workaround is to add the _SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING to the project defines as instructed, until the namespace can be removed (by upgrading gest?)

_Carmen Edit_: we should just upgrade gtest, as we are using a version that is quite a few years old now.

har00n-haider:
This fixes #435

## Description
- Updated gtest to 1.11 (Latest)
- Made a slight change to once class to fix new compilation error in bouncer test

## Test cases
 - Waiting on travis

### DIFF
--- a/bouncer/src/repo/core/model/bson/repo_bson.h
+++ b/bouncer/src/repo/core/model/bson/repo_bson.h
@@ -54,7 +54,7 @@ namespace repo {
 			//TODO: Eventually we should inherit from a generic BSON object.
 			//work seems to have been started in here:https://github.com/jbenet/bson-cpp
 			//alternatively we can use a c++ wrapper on https://github.com/mongodb/libbson
-			class REPO_API_EXPORT RepoBSON : private mongo::BSONObj
+			class REPO_API_EXPORT RepoBSON : public mongo::BSONObj
 			{
 				friend class RepoBSONBuilder;
 				friend class repo::core::handler::MongoDatabaseHandler;
--- a/bouncer/src/repo/core/model/bson/repo_bson.h
+++ b/bouncer/src/repo/core/model/bson/repo_bson.h
@@ -54,7 +54,7 @@ namespace repo {
 			//TODO: Eventually we should inherit from a generic BSON object.
 			//work seems to have been started in here:https://github.com/jbenet/bson-cpp
 			//alternatively we can use a c++ wrapper on https://github.com/mongodb/libbson
-			class REPO_API_EXPORT RepoBSON : public mongo::BSONObj
+			class REPO_API_EXPORT RepoBSON : private mongo::BSONObj
 			{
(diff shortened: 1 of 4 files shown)


bouncer/src/repo/core/model/bson/repo_bson.h
```diff
@@ -54,7 +54,7 @@ namespace repo {
 			//TODO: Eventually we should inherit from a generic BSON object.
 			//work seems to have been started in here:https://github.com/jbenet/bson-cpp
 			//alternatively we can use a c++ wrapper on https://github.com/mongodb/libbson
-			class REPO_API_EXPORT RepoBSON : private mongo::BSONObj
+			class REPO_API_EXPORT RepoBSON : public mongo::BSONObj
```
carmenfan, line 57: @har00n-haider  this is private on purpose - so we don't use the functions from the parents.

Where is it complaining?

har00n-haider: The full error is the commit message here:

https://github.com/3drepo/3drepobouncer/pull/546/commits/bf7ce11b63bf64df4bc0c5dfa648956188fa8e52

har00n-haider: I feel like it might be because we are printing the BSON somewhere using gtests print functions, I'll try narrowing it down

carmenfan: sounds like we're iterating on it somewhere - can we fix it?

(If you ran out of time don't worry)

har00n-haider: I could not get to a full solution, but this is what I have found out:

- There is a default template print function that is getting called by our test code 
- Using gtest macros like ASSERT_EQ require a method to print different types to an outputstream
- we don't have a 'foo::PrintTo(const T&, ostream*)' or 'operator<<(ostream&, const T&)' defined for the repoBson object (as required gtest functions [here](https://github.com/google/googletest/blob/100f6fbf5f81a82d163c1e29735e8a2936eacd4f/googletest/include/gtest/gtest-printers.h#L43))
- gtest suite uses its own template function, which causes the compilation error

I've pushed a commit to revert the change to the inhertance accessor, hopefully this is enough to debug the issue!

PR #577, opened 2022-05-18
Issue 576 remove writeConcern as a config option

### LINKED ISSUE #576: mongoimport should not run with a bespoke write concern specification
### Description
this addresses https://github.com/3drepo/DevOps/issues/410

also include a utility script to fix any broken toy projects

carmenfan:
This fixes #576

#### Description
- We no longer allow specific write concern configuration on mongoimport 
- Added a script to patch any toy imports that requires their data renamed due to this bug:
>   yarn run-rename --teamspace [teamspace name] --modelId [model id]

#### Test cases
- should now work in mongo v5 replicasets


keifgwinn (approved):
looks good to me

PR #594, opened 2022-12-13
ISSUE 592 - DGN Upload Bug

### LINKED ISSUE #592: DGN upload bug (ODA Upgrade)
### Description

Models uploading via DGN with  multiple views defined in the file are failing to process 

- This is happening in: both 
- This requires a hotfix to production: maybe 

### Steps to replicate
- upload a file - samples here - https://www.dropbox.com/sh/h1uedpj9akrjz1y/AAAiBD_tNTL35L4uUvbQ6Avea?dl=0 
- model will fail with a system error 
- Issue can be resolved by removing some views but will result in a model with no zoom capability? example - https://staging.dev.3drepo.io/viewer/Balfour_Beatty_TEST/6cce8840-6cd2-11ed-beff-055977518cdd 

### Current Behaviour
- models fail 

### Expected Behaviour
- models dont fail

sebjf:
This fixes #592 

#### Description
For now, this PR just updates ODA to 2023.11, which allows the new DGNs noted in 592 to import correctly.

Note that this does *not* allow 109968 to import, because the issue with that is that the nominated view does not have any geometry.


carmenfan (changes requested):
#### Missing materials on revit models
The green bit is clearly missing... is this to do with not getting the materials anymore or it's failing to grab the texture or something else?
left is staging (old ODA) right is this branch ([file here](https://www.dropbox.com/s/389jw9id1iai0no/rac_basic_sample_project.rvt?dl=0)): 
image

carmenfan:
@sebjf  I've updated travis to use the new teigha version but it's still failing to compile, can you take a look?

looks like the library is now using C++14?

sebjf:
Missing materials on revit models

This is indeed a missing texture. The `determineTexturePath` method fails when used with 23.11, however the paths seem identical as returned from ODA.

It seems the failure is in [this line](https://github.com/3drepo/3drepobouncer/blob/58013165a9deafcd4586acbe97472bd12bad2b3f/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_rvt.cpp#L63), which returns an empty string in the 23.11 build instead of the path representation of the relative texture path in 22.11, which is peculiar because we haven't changed the version of Boost...

Before:

image

Now:

image

If we modify this snippet to store the path reference before calling `make_preferred()`, the method functions correctly again.

Once the textures are fixed though there is an issue with the tiling.

Before:

image

After:

image


This is due to the [OdGiMapper behaviour](https://github.com/3drepo/3drepobouncer/blob/58013165a9deafcd4586acbe97472bd12bad2b3f/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_rvt.cpp#L268) changing:

Before:

image

After:

image

carmenfan (approved):

carmenfan:
product team approved via https://github.com/3drepo/3drepo.io/pull/3947

PR #596, opened 2023-01-19
Issue #556 - avoid reporting on licensing errors

### LINKED ISSUE #556: Bouncer worker reporting failed unity licenses as 0 exit code successful elastic records
### Description

We shouldn't try and emit a bouncer log if we're in license retry state, this lead to 17k of bad records when left on retry loop for a few days.

- This is happening in: prod/staging/both 
- This requires a hotfix to production: no 

### Steps to replicate
- Configure for elastic
- Spin up new bouncer
- Do not license unity
- submit job

### Current Behaviour
- every fail reported as a success

### Expected Behaviour
- do not emit a record for license retries.

keifgwinn:
This fixes #556

#### Description
- decision to report process usage now waits to see if we're licensed correctly and if not just logs locally.

#### Test cases
- logs shouldn't be sent to elastic if we're not licensed.


keifgwinn:
does it make sense to change this to the RevisionID rather than ModelID for the management of reports @carmenfan ?

carmenfan:
> does it make sense to change this to the RevisionID rather than ModelID for the management of reports @carmenfan ?

Change what? Not quite catching the context 😆

keifgwinn:
so I updated it to move the data from the PID to a new structure based on the modelid (because the rid wasn't being sent to elastic before) and it's the key that allows it to find the data after the process has finished. Would it be possible for two revisions on the same model to be in flight in the same bouncer instance, or perhaps unity and modelq at the same time. Not in the way that we have it set up in prod, but on single instances it could happen

carmenfan:
> Would it be possible for two revisions on the same model to be in flight in the same bouncer instance

how it's deployed aside, on the code level it's possible as we don't block users pushing more than 1 revision to process on the API

So yes Model Id would not be the most reliable thing to use if you want to uniquely identify a job, rId is much better

tools/bouncer_worker/src/lib/processMonitor.js
```diff
@@ -10,6 +10,7 @@ const ProcessMonitor = {};
 const logLabel = { label: 'PROCESSMONITOR' };
 
 const dataByPid = {};
+const dataByModel = {};
```
carmenfan, line 13: it's probably better to call this `dataByRid`

carmenfan: do you really need both dataByPid and dataByModel? 

I think you could just reference everything by `dataByModel`?

tools/bouncer_worker/src/lib/processMonitor.js
```diff
@@ -51,17 +52,21 @@ const updateMemory = async (pid) => {
...
+		}
 	}
 };
 
 const monitor = (pid) => setInterval(() => updateMemory(pid), memoryIntervalMS);
 
 const shouldMonitor = async () => enabled && permittedOS.includes(await currentOSPromise);
-const monitorEnabled = shouldMonitor();
+
+ProcessMonitor.enabled = shouldMonitor();
```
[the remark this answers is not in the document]
keifgwinn: ok, made those changes

tools/bouncer_worker/src/lib/processMonitor.js
keifgwinn: I think we do, the code that executes the command only knows about it's PID and the code that sends the report only knows about the RID, so simplest to leave the PID for active process, RID for after the process is completed.

carmenfan: From what I can see we don't actually use the Pid inside the processMonitor (updateMemory is called by Monitor which could've been called with the `rId`: 

image

the other place that reference pid is the call to `stopMonitor`, which you can call inside with rId as it has reference to `processInformation` (which is how you construct the `dataByRid` with at `startMonitor`

image


So unless I'm missing something (which I often am!), you should be able to just `dataByRid` for everything?

tools/bouncer_worker/src/lib/processMonitor.js
```diff
@@ -51,17 +52,19 @@ const updateMemory = async (pid) => {
...
 
 const monitor = (pid) => setInterval(() => updateMemory(pid), memoryIntervalMS);
 
 const shouldMonitor = async () => enabled && permittedOS.includes(await currentOSPromise);
-const monitorEnabled = shouldMonitor();
 
 ProcessMonitor.startMonitor = async (startPID, processInfo) => {
-	if (!(await monitorEnabled)) return;
+	if (!(await shouldMonitor())) return;
+	if (dataByRid[processInfo.model]) delete dataByRid[processInfo.model];
```
carmenfan, line 67: Think we forgot to change it here

```suggestion
	if (dataByRid[processInfo.Rid]) delete dataByRid[processInfo.Rid];
```

carmenfan (changes requested):
Also see the extra comment in the `dataByPid` conversation

tools/bouncer_worker/src/lib/processMonitor.js
keifgwinn: yes, I had some kind of mental block on thinking it needed to know the accurate PID to monitor the mem utilisation, but as we're just checking the waterlevel of total system we don't need pid.

PR #601, opened 2023-02-20
ISSUE #600 keep all data in metres in an IFC import

### LINKED ISSUE #600: export IFC models in m
Disable the use original units flag so all IFC exports will export in metres

carmenfan:
This fixes #600

#### Description
unflag CONVERT_BACK_UNITS

#### Test cases
all models should now export in metres


carmenfan:
switching to merge into #611 as the requirements have evolved

PR #624, opened 2023-05-06
ISSUE #623 construct the supermesh cache file in bouncer

### LINKED ISSUE #623: Merge all supermesh metadata into a single file
### Description
extension of problem explained here

https://github.com/3drepo/3drepo.io/issues/4212

To get around this on new models, we should just save a copy of all the json mappings as a single file

Old json mappings are also saved for streaming milestone 2 purposes.

carmenfan:
This fixes #623

#### Description
2nd part of the solution to https://github.com/3drepo/3drepo.io/issues/4212
- aggregate all unity json mappings into a single and store it as `<rev_id>/supermeshes.json`

#### Test cases
- should work as expected
- 3drepo.io should no longer need to generate this file on the fly for new model uploads

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -131,7 +131,7 @@ matrix:
           - export OCCT_ROOT=$PWD/OCCT
           - export IFCOPENSHELL_ROOT=$PWD/IfcOpenShell
           - echo ============ TEIGHA INSTALL =============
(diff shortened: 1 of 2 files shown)


sebjf (approved):
All working and I can't see any regression issues.

PR #656, opened 2023-12-06
Issue 654 - gzipping asset bundles 

### LINKED ISSUE #654: Model Loading Speed Improvements
### Description

Bouncer issue for model loading speed improvements, now we know that at least some change will be made to bouncer. 

Tasks and Goals to follow.

Product issue with investigation and discussion: https://github.com/3drepo/3D-Repo-Product-Team/issues/346

sebjf:
This fixes #654

#### Description
This PR gzips the Unity Asset Bundles by default, and provides a metadata BSON so that the state of this encoding is stored in the db.

This PR exists for now just for feedback.


sebjf:
Hi @Charence, the build failed on this branch previously. I have started it again but both times I cannot see its progress:

image

Do you know what's up with it?

wrapper/src/c_sharp_wrapper.cpp
```diff
@@ -386,13 +390,24 @@ bool CSharpWrapper::saveAssetBundles(
...
+			std::ostringstream ss; // memory stream containing the binary data to write to the database
+
+			// All unity asset bundles are gzipped by default
+
+			boost::iostreams::filtering_streambuf<boost::iostreams::input> in;
+			in.push(boost::iostreams::gzip_compressor());
+			in.push(inputStream);
+			boost::iostreams::copy(in, ss);
+
+			repo::core::model::RepoBSONBuilder builder;
```
carmenfan, line 406: @sebjf 

I wouldn't do it this way. This is quite difficult to read and making the way we store metadata data about a file highly coupled with the functions/a classes that are simply passing in the files.

I think this could be a lot simpler (Unless i'm missing something, which I'm often am!). I think the file manager can manage the gzipping and it should just be a flag on the function https://github.com/3drepo/3drepobouncer/blob/master/bouncer/src/repo/core/handler/fileservice/repo_file_manager.h#L60

If this is set to true, it would know to append the gzip flag onto the metadata of the file ref.

We could just gzip all the geo files we export, or it could be a flag in the exporter to indicate whether we want to gzip this. (Persumably we benefit from gzipping the SRCs too, but I'm not sure how unreal handles those requests!)

And it would also make gzipping the JSON caches much easier to do if we want to implement it in the future.

carmenfan (changes requested):

wrapper/src/c_sharp_wrapper.cpp
sebjf: Hi @carmenfan, this should be resolved now!

carmenfan (approved):

carmenfan (changes requested):
image

@sebjf 
This runs fine, works fine, but failing to build in linux

sebjf:
> image
> 
> @sebjf This runs fine, works fine, but failing to build in linux

@carmenfan, this should be fixed now for when Travis comes back!

PR #660, opened 2023-12-26
Issue 608 Provide better reasoning in the log for failures to write to file system

### LINKED ISSUE #608: Provide better reasoning in the log for failures to write to file system
### Description

We're seeing a lot more failed to write to filesystem since v5, with a fixed retry of only a few seconds. However, I've noticed this across all the environments we're running in now, local, AWS efs, on premise NAS, and so I suspect there might be something else happening than actually something stopping the writes to disk.

I realise this needs more investigation, but I'm opening this ticket to look into it so I don't forget to mention it.

I've seen this behaviour in prod, on staging, and on local.

It would be good to be able to make this timeout/retry configurable to aid in investigations, and also treat FS write failures as soft fails worthy of returning to the queue.

- This is happening in: both 
- This requires a hotfix to production: yes/no 

Edit by CF: this is where we're writing the file. We need to investigate if we can catch the error better and find out the reason and dump it into the logs https://github.com/3drepo/3drepobouncer/blob/master/bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp#L138

### Steps to replicate
- TBD

### Current Behaviour
- Occasional file system errors that don't appear to be file system related as every time I've tested immediately on it it's been able to write.

### Expected Behaviour
- no errors

sbanethia:
…to file system

Added functions to check the file writing permissions and also check the free disk size before writing the file on the disk.

This fixes #608

#### Description
Added functions to check the file writing permission and also to check if the size on the disk is free or not to get proper error descriptions in the log.

### DIFF
--- a/bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
+++ b/bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
@@ -133,15 +133,37 @@ std::string FSFileHandler::uploadFile(
 	path /= keyName;
 	ss << keyName;
 	int retries = 0;
-	bool failed;
+	bool failed = false;
 	do {
 		std::ofstream outs(path.string(), std::ios::out | std::ios::binary);
-		outs.write((char*)bin.data(), bin.size());
-		outs.close();
-		if (failed = (!outs || !repo::lib::doesFileExist(path))) {
-			repoError << "Failed to write to file " << path.string() << ((retries + 1) < 3 ? ". Retrying... " : "");
+		if (outs.good()) 
+		{
+			outs.write((char*)bin.data(), bin.size());
+			outs.close();
+		}
+		else if(outs.bad() || outs.fail()) 
+		{
+			failed = true;
+			if (repo::lib::doesFileExist(path)) {
+				repoError << "Failed to write to file " << path.filename() <<" as file is already present. ";
+			}
+
+			if (repo::lib::hasFileWritePermissions(path)) {
+				repoError << "Failed to write to file " << path.filename() << " as file writting permission is not there.";
+			}
+
+			if (repo::lib::isFileWritingSpaceAvailable(path)) {
+				repoError << "Failed to write to file " << path.filename() << " as there is insufficient space on the disk.";
+			}
+
+			repoError << ((retries + 1) < 3 ? ".Retrying... " : "");
 			boost::this_thread::sleep(boost::posix_time::seconds(5));
 		}
+		
+		/*if (failed = (!outs || !repo::lib::doesFileExist(path))) {
+			repoError << "Failed to write to file " << path.string() << ((retries + 1) < 3 ? ". Retrying... " : "");
+			boost::this_thread::sleep(boost::posix_time::seconds(5));
+		}*/
 	} while (failed && ++retries < 3);
 
 	return failed ? "" : ss.str();
--- a/bouncer/src/repo/lib/repo_utils.h
+++ b/bouncer/src/repo/lib/repo_utils.h
@@ -60,5 +60,25 @@ namespace repo {
 			char* value = getenv(envVarName.c_str());
 			return (value && strlen(value) > 0) ? value : "";
 		}
+
+		static bool hasFileWritePermissions(const boost::filesystem::path& inputPath) 
+		{
+		
+			boost::filesystem::file_status s = boost::filesystem::status(inputPath);
+			if (boost::filesystem::perms::owner_write == s.permissions()) {
+				return true;
+			}
+			return false;
+		}
+
+		static bool isFileWritingSpaceAvailable(const boost::filesystem::path& inputPath) 
+		{
+			boost::filesystem::path pathWithoutFileName = inputPath;
+			boost::filesystem::space_info spaceInfo = boost::filesystem::space(pathWithoutFileName.remove_filename());
+			if (spaceInfo.free > (boost::filesystem::file_size(inputPath) + 1)) {
+				return true;
+			}
+			return false;
+		}
 	}
 }


bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
```diff
@@ -133,15 +133,37 @@ std::string FSFileHandler::uploadFile(
...
-			repoError << "Failed to write to file " << path.string() << ((retries + 1) < 3 ? ". Retrying... " : "");
+		if (outs.good()) 
+		{
+			outs.write((char*)bin.data(), bin.size());
+			outs.close();
+		}
+		else if(outs.bad() || outs.fail()) 
+		{
+			failed = true;
+			if (repo::lib::doesFileExist(path)) {
```
carmenfan, line 147: I don't think this is an issue? if we're overwriting an existing file we are overwriting an existing file. it wouldn't be an error?

bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
```diff
@@ -133,15 +133,37 @@ std::string FSFileHandler::uploadFile(
...
+	bool failed = false;
 	do {
 		std::ofstream outs(path.string(), std::ios::out | std::ios::binary);
-		outs.write((char*)bin.data(), bin.size());
-		outs.close();
-		if (failed = (!outs || !repo::lib::doesFileExist(path))) {
-			repoError << "Failed to write to file " << path.string() << ((retries + 1) < 3 ? ". Retrying... " : "");
+		if (outs.good()) 
+		{
+			outs.write((char*)bin.data(), bin.size());
```
carmenfan, line 141: as it stands we're no longer detecting error on the .write().

bouncer/src/repo/lib/repo_utils.h
```diff
@@ -60,5 +60,25 @@ namespace repo {
 			char* value = getenv(envVarName.c_str());
 			return (value && strlen(value) > 0) ? value : "";
 		}
+
```
carmenfan, line 63: Is there no way to detect these error reasons from the write attempt instead of preemptively check for these?

The whole point of this is to try and print the error if possible, if it's not possible I'm not sure if we should bother with this issue.

carmenfan (changes requested):

bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
```diff
@@ -133,15 +133,37 @@ std::string FSFileHandler::uploadFile(
...
+
+			if (repo::lib::isFileWritingSpaceAvailable(path)) {
+				repoError << "Failed to write to file " << path.filename() << " as there is insufficient space on the disk.";
+			}
+
+			repoError << ((retries + 1) < 3 ? ".Retrying... " : "");
 			boost::this_thread::sleep(boost::posix_time::seconds(5));
 		}
+		
+		/*if (failed = (!outs || !repo::lib::doesFileExist(path))) {
```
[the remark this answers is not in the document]
sbanethia: Done

bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
sbanethia: The above code is reverted back as per the discussion.

bouncer/src/repo/lib/repo_utils.h
sbanethia: The above code is reverted back as per the discussion.

bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
sbanethia: The above code is reverted back as per the discussion.

bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
```diff
@@ -136,8 +136,17 @@ std::string FSFileHandler::uploadFile(
...
 	do {
 		std::ofstream outs(path.string(), std::ios::out | std::ios::binary);
-		outs.write((char*)bin.data(), bin.size());
-		outs.close();
+
+		try {
+			outs.write((char*)bin.data(), bin.size());
+			outs.close();
+		}
+		catch (const boost::filesystem::filesystem_error& e)
```
carmenfan, line 144: Does this work? 

Considering we're using the std ofstream here, I would've never dreamt of it throwing a boost error - or am I missing something?

bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
```diff
@@ -112,17 +112,17 @@ std::vector<std::string> FSFileHandler::determineHierachy(
 }
 
 std::string FSFileHandler::uploadFile(
-	const std::string          &database,
-	const std::string          &collection,
-	const std::string          &keyName,
-	const std::vector<uint8_t> &bin
+	const std::string& database,
```
carmenfan, line 115: Are you using a code cleaner that is doing this?

Can we revert? Imo & should be closer to the variable (also you're making it inconsistent with the rest of the code base)

This applies to the other PR as well.

carmenfan (changes requested):

bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
sbanethia: As per the suggestion I have turned off auto format in my visual studio due to which extra space was added in the code.

bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
sbanethia: We have changed the file writing class from std::ofstream to boost::filesystem::ofsream for better error handling and log writing on getting an exception.

carmenfan: Have you tested it to see if it actually throws an exception? Looking at the boost documentation it claims it has no semantic differences, so it's not going to be doing anything that std library version doesn't do.

image

carmenfan (changes requested):
Just a question on whether this would achieve what we're trying to do.

If it's not possible - we may as well close the issue

bouncer/src/repo/core/handler/fileservice/repo_file_handler_fs.cpp
sbanethia: @carmenfan We can do the exception handling but it is not giving proper error in a proper readable format, so this is not useful.

PR #677, opened 2024-02-16
Issue 666 - add support for RepoBundles

### LINKED ISSUE #666: Add support for AssetGenerator
### Description

We want to [replace Unity AssetBundles with our own transport format](https://github.com/3drepo/3drepounity/issues/501) that will bypass the WASM VM, and load faster. These assets will be built by a separate library.

### Goals

- [x] [AssetGenerator](https://github.com/3drepo/AssetGenerator) is an optional dependency 
- [x] Control is based on a CMake argument
- [x] Bundles are built using the [AssetGenerator](https://github.com/3drepo/AssetGenerator) library and written to the database

### Tasks

- [x] Add AssetGenerator as a git submodule
- [x] Update AssetsList so that it can support different assets

### Related Resources

Product Issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/376

sebjf:
This fixes #666 - Add support for RepoBundles

#### Description
This PR contains mainly minor changes required to integrate [AssetGenerator](https://github.com/3drepo/AssetGenerator). 

AssetGenerator is an optional dependency that can be turned off (turned on by default) with the `REPO_ASSETGENERATOR_SUPPORT` CMake flag. AssetGenerator will create `RepoBundles`, which are documented its own repository, after the stash scene is created.

The changes made to bouncer to support this are:

1. The cmake project and build scripts have been updated to include AssetGenerator as a git submodule
2. A new attribute array, ids, has been added to `MeshNode`
3. The unityAssets.json schema has been extended, and it now updates instead of replacing the document in the database.
4. The `_RepoVector3D` constructor has been adjusted so it can be initialised by value (it makes a copy of the reference immediately anyway), and the type has been given two new members.
5. `repo_color4d_t` has been given new members.
6. `PropertyTree::addToTree` has support for `repo_color4d_t` added.
7. Assimp no longer initialises unknowns to nans, as these are used in calculations which can result in invalid json.


bouncer/src/repo/manipulator/modelutility/repo_scene_manager.cpp
```diff
@@ -69,16 +74,17 @@ bool SceneManager::commitWebBuffers(
 	}
 
 	std::string errMsg;
-	if (REPO_COLLECTION_STASH_UNITY == geoStashExt)
+
+	if(!resultBuffers.unityAssets.isEmpty())
 	{
-		if (success &= handler->upsertDocument(databaseName, projectName + "." + geoStashExt, resultBuffers.unityAssets,
-			true, errMsg))
+		if (success &= handler->upsertDocument(databaseName, projectName + "." + assetsStashExt, resultBuffers.unityAssets,
```
carmenfan, line 80: any reason why we set the overwrite flag to false here?

In normal usage we don't ever anticipate to need to overwrite this, but in the event where we want to re-generate the stash for some reason, this would fail without the overwrite flag

bouncer/src/repo/core/model/bson/repo_bson_factory.h
```diff
@@ -265,12 +267,13 @@ namespace repo {
 				* @return returns a mesh node
 				*/
 				static MeshNode makeMeshNode(
```
carmenfan, line 269: 2 thoughts here:

1. I think we should separate MeshNode and SuperMesh node. They could converge to a single private function in the cpp of this file, but I think it'll give better readability and less confusion about what this id array is suppose to represent.
2. We should rename `ids` to subMeshIds` or something. as it definitely confused me when I read this code 😆

bouncer/src/repo/core/model/bson/repo_bson_factory.h
```diff
@@ -152,14 +152,16 @@ namespace repo {
 				*/
 				static RepoUnityAssets makeRepoUnityAssets(
 					const repo::lib::RepoUUID                              &revisionID,
-					const std::vector<std::string>                         &assets,
+					const std::vector<std::string>                         &unityAssetFiles,
```
carmenfan, line 155: Is this a temporary measure (i.e. will be split when we move to v5 endpoint)? I think we should keep the 2 mapping separate asset mapping separate, considering they have different json and binary files they don't really share anything in common.

bouncer/src/repo/core/model/bson/repo_node_mesh.cpp
```diff
@@ -265,6 +265,17 @@ std::vector<repo_color4d_t> MeshNode::getColors() const
 	return colors;
 }
 
+std::vector<float> MeshNode::getIds() const
```
carmenfan, line 268: as above, perhaps we should rename to getSubmeshIds

bouncer/src/repo/core/model/bson/repo_node_mesh.h
```diff
@@ -49,6 +49,7 @@ namespace repo {
 #define REPO_NODE_MESH_LABEL_UV_CHANNELS_BYTE_COUNT	"uv_channels_byte_count"
 #define REPO_NODE_MESH_LABEL_SHA256                  "sha256"
 #define REPO_NODE_MESH_LABEL_COLORS                  "colors"
+#define REPO_NODE_MESH_LABEL_IDS                     "ids"
```
carmenfan, line 52: as above

bouncer/src/repo/manipulator/modelutility/repo_mesh_map_reorganiser.cpp
```diff
@@ -157,7 +157,14 @@ repo::core::model::MeshNode MeshMapReorganiser::getRemappedMesh() const
 			bboxArr.push_back({ bbox[1].x, bbox[1].y, bbox[1].z });
 		}
 
-		auto newMesh = repo::core::model::RepoBSONFactory::makeMeshNode(newVertices, newFaces, newNormals, bboxArr, newUVs, newColors);
+		auto submeshIds = getIDMapArrays();
```
carmenfan, line 160: Not something to address here, just noting down my thoughts.

Stash graph and this should probably merge into 1 in the future? We're effectively batching the meshes twice 😆

bouncer/src/repo/manipulator/modelutility/repo_mesh_map_reorganiser.cpp
```diff
@@ -157,7 +157,14 @@ repo::core::model::MeshNode MeshMapReorganiser::getRemappedMesh() const
...
+		std::vector<float> newIds;
+		for (auto& buf : submeshIds)
+		{
+			newIds.insert(newIds.end(), buf.begin(), buf.end());
+		}
+
+		auto newMesh = repo::core::model::RepoBSONFactory::makeMeshNode(newVertices, newFaces, newNormals, bboxArr, newUVs, newColors, newIds);
```
carmenfan, line 167: do we really need this id array?

Is this not the same as the ids within the meshMapping?

bouncer/src/repo/lib/datastructure/repo_vector3d.h
```diff
@@ -24,15 +24,19 @@
 #include <sstream>
 #include <cmath>
 
+// Undefine windef.h's wierdness
```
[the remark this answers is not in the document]
sebjf: Unfortunately not; the difference is that `NOMINMAX` can prevent windef.h making the defines in the first place, but repo_vector3d.h is often included after something that brings in windef.h, in which case it's too late.

The alternative is to find everywhere repo_vector3d.h is used, and move it to the top. I chose undef over this, because that would touch a load of unrelated files, and also mean the ordering of the headers would become significant without there being any indication that was the case, leading to a bunch of obscure errors if someone reorders them, adds a new header, includes repo_vector somewhere new, etc..

bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_asset.cpp
```diff
@@ -233,19 +233,22 @@ bool AssetModelExport::generateTreeRepresentation()
 			superMeshesTree.addToTree("model", scene->getProjectName());
 			superMeshesTree.addArrayObjects("supermeshes", meshesTrees);
 
-			jsonTrees["/revision/" + scene->getRevisionID().toString() + "/supermeshes.json"] = superMeshesTree;
+			jsonTrees["/revision/" + scene->getRevisionID().toString() + "/supermeshes_unity.json"] = superMeshesTree;
```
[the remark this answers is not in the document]
sebjf: It shouldn't: this file (`supermeshes_unity.json`) is the container-wide cache file (not the regular mappings file). 

In "normal" legacy mode (i.e existing revisions), they will request `supermeshes.json`.

`supermeshes_unity.json` should only be read if the user is operating in forced-legacy mode. In this case yes it will indeed not exist for existing revisions, but .io will re-create it on demand, as it did for models before we added the cache.

The `generateSuperMeshMappings` function in .io has not changed, since all post-RepoBundles containers should be imported with the cache files already existing, and the existing unityAssets.json should be backwards-compatible.

(The idea is that in a couple of releases when it goes away, we are left with just one API that works with new and existing cache files transparently, without having to know which type of bundle is underneath.)

bouncer/src/repo/core/model/bson/repo_bson_factory.h
sebjf: It is not intended to be - the idea is that having one unityAssets.json, the viewer doesn't need to know in advance which type of bundles it should be asking for.

The other option would for bouncer to write two separate documents, and for .io to return the legacy one if the new one doesn't exist.
This would in some ways be neater because we could add a new API, and the existing one is left for forced-legacy mode.
In this case the viewer would rely on both documents having the same schema, and on the file extension to figure out which type of bundle it should request.

If you'd like me to refactor it to operate like that? Or something else if you have another idea!

CMakeLists.txt
```diff
@@ -149,6 +149,7 @@ option (REPO_BUILD_CLIENT "If the Command Line Client is built in addition to th
 option (REPO_BUILD_TOOLS "If the Command Line Tool is built in addition to the library" OFF)
 option (REPO_BUILD_TESTS "If the test suite for the core bouncer logic is built in addition to the library" OFF)
 option (REPO_NO_GZIP "If zlib is not compiled into boost" OFF)
+option (REPO_ASSETGENERATOR_SUPPORT "If the AssetGenerator is present and compiled into the library" ON)
```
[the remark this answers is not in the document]
sebjf: Done!

bouncer/src/repo/core/model/bson/repo_bson_factory.h
carmenfan: > This would in some ways be neater because we could add a new API, and the existing one is left for forced-legacy mode.

I think that' would be a neater approach 👍

bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_asset.cpp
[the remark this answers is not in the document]
carmenfan: ok I don't think I fully understand this yet... 😆 

so supermeshes_unity.json store different information to the old supermeshes.json and the latter is no longer used with repo bundles, and we no longer generate the old `supermeshes.json`?

So this becomes an issue if we have to rollback, with new models it will try to load supermeshes.json which doesn't exist and 3drepo.io will have to generate the cache on the fly?

(the 3drepo.io generate cache is never ran live - it was there so I can run it through a script 🙈 the problem was that this cache was taking upwards of 6mins to run so it becomes a problem)

I suppose there's no easy way to write both for now and remove the old one in the future?

bouncer/src/repo/lib/datastructure/repo_vector3d.h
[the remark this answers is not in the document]
carmenfan: ok fair enough. 

In which case can we please fix the spelling on `weird` 🙈

bouncer/src/repo/manipulator/modelconvertor/export/repo_model_export_asset.cpp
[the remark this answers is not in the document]
sebjf: > (the 3drepo.io generate cache is never ran live - it was kept there so I can run it through a script 🙈 the problem was that this cache was taking upwards of 6mins to run so it becomes a problem)

OK I didn't realise that!

> so supermeshes_unity.json store different information to the old supermeshes.json and the latter is no longer used with repo bundles, and we no longer generate the old supermeshes.json?

No I haven't explained it well 😄 

The supermeshes.json (cache) continues to be generated and used by RepoBundles, just like it was for AssetBundles. Whereas before it was built by `AssetModelExport`, its now built by `RepoBundleExport`. There is only one file and one API for new and existing imports, and the viewer handles the existing versions transparently.

This is possible because the submeshid layouts are identical between AssetBundles and RepoBundles, so the same Mappings files apply to both. The only difference is the old ones don't have additional material data the `RepoBundles` expect.

Therefore, the idea going forward is that the viewer has one pathway for handling the mappings, and only the Bundles themselves have dedicated import functions.

This is for maintainability. The function to handle the mappings is quite complex, and the only difference between old and new is the additional bit of logic to handle materials. Further, when we get to Milestone 2 proper, its likely the logic in this area will undergo significant refactoring, so we ideally don't want to indefinitely look after two paths that are 95% identical.

In fact, the original implementation had just one set of Jsons in unityAssets.

However, I split them out in unityAssets, and added the supermeshes_unity.json above, so that bouncer would continue to generate complete mappings independently of RepoBundles, in case there turned out to be an issue.

I can't think of what that might be, but then if I could be 100% sure there are no problems with RepoBundles we wouldn't need to run them side by side for a time!

In brief then, this file is a backup for a backup: if there's a problem with RepoBundles we should in theory be able to fall back to AssetBundles + new mappings. However, if there is a problem with RepoBundles AND the new mappings, we can fallback to AssetBundles + supermeshes_unity.json.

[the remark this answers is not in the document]
carmenfan: ok, fair enough 👍

bouncer/src/repo/lib/datastructure/repo_vector3d.h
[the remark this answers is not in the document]
sebjf: 🙈 fixed!

bouncer/src/repo/lib/datastructure/repo_vector3d.h
```diff
@@ -24,15 +24,19 @@
...
 namespace repo{
 	namespace lib{
 		template <class T>
 		class  REPO_API_EXPORT _RepoVector3D
 		{
 		public:
 
-			_RepoVector3D<T>(const T &x = 0, const T &y = 0, const T &z = 0) : x(x), y(y), z(z) {}
-			_RepoVector3D<T>(const std::vector<T > &v)
+			_RepoVector3D<T>(T x = 0, T y = 0, T z = 0) : x(x), y(y), z(z) {}
```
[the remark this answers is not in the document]
sebjf: Yes theres no need, added them back!

bouncer/src/repo/manipulator/modelutility/repo_mesh_map_reorganiser.cpp
carmenfan: @sebjf 

I guess my issue here is that we're blurring the line between what we declare as a mesh data within the BSON document and what we're using as an interm structure for the mesh batching 🤔 

MeshNode within RepoBsonFactory is suppose to be a 1:1 mapping with the schema that is stored in the database. We "augmented" this for the super meshes which we still store as a mesh in stash.3drepo as a scene structure.

The weird thing here is that we've augmented the bson class to add a property within the database schema, that never hits the database - which is a bit weird. Also the naming of `ids` is confusing as a mesh has `id`, but it's not immediately obvious what `ids` would mean in the context.

I guess what this should really be is a in memory structure that we store as a temporary thing as we construct the repo bundle mappings (that is not a clash of RepoBSONFactory)? Going forward, if we stop storing the stash graph then there isn't really much reason to have the meshMapping within the repo MeshNode even. It would probably be more memory/performance efficient as well, since by not piggy backing onto the mongo library it doesn't have silly enforcements such as immutable properties and storing everything into a serialised buffer etc....

I'm not sure what's the best way forward for this issue? whether we leave it as is and make it a note to refactor when we get rid of the stash graph (In which case I think we should rename ids to `subMeshIds` or something), or we just extrapolate the ids from the mesh mapping, or refactor this all together..

Whatever you feel is best 😆

sebjf: > MeshNode within RepoBsonFactory is suppose to be a 1:1 mapping with the schema that is stored in the database. 

Ah good point, I wasn't really looking at it from this perspective...

I am working on on splitting MeshNode/SupermeshNode as per comments above, so hopefully the discrepancy can be resolved in that.

The question is though is still how to do it. Semantically, the correct thing is to make SupermeshNode subclass MeshNode but then again we have a RepoNode type that never gets written to the database...

Another possibility is to define a new Supermesh type that contains everything MeshNode does,but is in-memory only, and have `getRemappedMesh()` return that?

Then RepoBundleExport can operate on this new in-memory type from then on.

MeshMapReorganiser would also be able to return a MeshNode for the purposes of AssetBundleCreator until it is decommissioned, after which supermeshes would only be handled in memory, and the mappings could be removed from MeshNode entirely?

carmenfan: @sebjf 

> The question is though is still how to do it. Semantically, the correct thing is to make SupermeshNode subclass MeshNode but then again we have a RepoNode type that never gets written to the database...

yea it's a bit of an overkill if we are not going to be storing any form of supermesh node soon enough..

> Another possibility is to define a new Supermesh type that contains everything MeshNode does,but is in-memory only, and have getRemappedMesh() return that?

possibly, that's kind of what we do during file imports when we're not quite ready to to construct the MeshNodes, so the behaviour would be similar to other parts of the code.

I think the in memory solution probably works best, but obviously the most rewrite. Not sure whether you feel like it would make more sense to do it when we scrap the stash graph or it makes sense to do it now. I leave this up to you 😄

sebjf: I think its best to split MeshNode, address #678 and decommission AssetBundleCreator & the stash graph in one ticket after RepoBundles have been tested for a while, as they are all interdependent.

My experience over the past few weeks is that trying to make changes while keeping AssetBundleCreator & associated APIs working has usually opened a can of worms, and I expect this to be the same!
If we do it once we don't have to worry about backwards compatibility it will go quicker, and with the increased flexibility will probably end up neater overall.

bouncer/src/repo/core/model/bson/repo_bson_factory.h
sebjf: Hi @carmenfan, I am thinking about how to support the two assets lists going forward. 
.io works by looking for the first document that matches the revision id in `stash.unity3d`.

The simplest approach is to make a new collection `stash.repobundles`, and then fallback to `stash.unity3d` if no documents are found. The downside is this adds another collection, and we need to explicitly check the old one in the fallback case indefinitely. 
It is by far the simplest case however and easy to understand.

For the future, it would be nice to have them in one collection with the same access mechanism, because then findOne will simply return the RepoBundles one for new imports and the Unity one for existing imports.
Though, for the period where we have both they need to be disambiguated/need different ids.

So far the only way I can think to do this is to take a similar approach for the `supermeshes_unity.json` fallback.

That is, have `AssetModelExport` start writing the Unity Assets document with a post-fixed name or some member to disambiguate it.

.io will `findOne` against `.unity3d`, returning the RepoBundle one where available, or the Unity one for existing imports. If the unity one is explicitly requested .io will search for that, then after a release hopefully, we simply stop writing the unity one.

Which approach would you prefer? (Or, do you have a better idea!)

carmenfan: could we do it so that they're completely separate endpoints, stored in different place. and let unity control the fall back?

i.e. the unity viewer asks for repo bundles asset list -> if response is 404, ask for the asset bundle list?

I'd prefer we don't blur the lines and do the fallback on 3drepo.io and make assumptions on what the user is asking for. especially if the schema might diverge in the future.

3drepo.io as a provider of data I think we want to be explicit, I give you exactly what you're asking for or nothing at all. Whether the client can support a backup option or not is the client's consideration. For e.g. if we moved unreal to use repo bundles or 3rd party integrators using this endpoint, they may not be able to interpret the unity bundles so we shouldn't confuse them

sebjf: @carmenfan yes sounds good! How do you want to handle the supermeshes.json (cache)? Should this be an exception or shall we create a permanent new API (e.g. `supermeshes_repo.json`)?

sebjf: Though what should we do about the assets list for federations that might mix new and existing imports? In this case .io would need to be aware of both because it only returns one response...

bouncer/src/repo/manipulator/modelutility/repo_scene_manager.cpp
sebjf: I have put this back since we have separate asset lists now!

bouncer/src/repo/core/model/bson/repo_bson_factory.h
sebjf: OK, as discussed, bouncer now writes the assets lists separately to `.stash.unity3d` and to `.stash.repobundles`. The v4 APIs implement backwards compatibility on .io for now, which will be decommissioned when the viewer moves to v5 style APIs and takes over backwards compatibility on a per-container basis.

bouncer/src/repo/core/model/bson/repo_node_mesh.h
sebjf: Done!

bouncer/src/repo/core/model/bson/repo_node_mesh.cpp
sebjf: Done!

bouncer/src/repo/core/model/bson/repo_bson_factory.h
sebjf: Done!

carmenfan (approved):

PR #681, opened 2024-05-02
Issue 667

### LINKED ISSUE #667: A new error code for model import failures due to encrypted file
### Description
Occasionally we get files (we've seen it on nwds, dwgs) that are encrypted and thus cannot import:
image

These are currently returning error code 5, but we should make a new error code to show it's a problem user can address.

### Goals
- [ ] As a user I would like to be informed if the cause of an import failure is because the file is encrypted.

### Tasks
TBC

sebjf:
This fixes #667 - new error code for encrypted NWD files

#### Description
This PR adds an extra case to the NWD File Processor that detects the appropriate OdError code when an encrypted file is provided, and returns the corresponding 3D Repo error code (37).

This PR relies on #680, which should be merged first. This PR exists at the moment just to indicate the issue has been addressed.

#### Test cases
A new unit test has been added with an encrypted NWD.


carmenfan (approved):

carmenfan (approved):
Looks fine, merging with #680

PR #684, opened 2024-05-13
Issue 683 - clear the default graph as soon as possible

### LINKED ISSUE #683: bouncer should free the graph after the stash graph has been created
### Description

After creating the stash graph, bouncer should free any unused resources to limit peak memory.

This is in response to a file that is exceeding the 26 Gb limit of the Pods due to the number of instances.

This should be a straightforward case of implementing a default graph equivalent of `clearStash` and calling it in `commitScene`, since the graphs should never be used at the same time. Though, we probably want to refactor all references to graph in RepoScene, to be sure that nothing attempts to load an unloaded default graph.

### Goals

- [ ] Default graph is freed before RepoBundles are created.

### Tasks

- [ ] Implement `clearDefault()` on `RepoScene`
- [ ] Call after the selection tree is generated
- [ ] Add checks to detect if the default tree has been unloaded

### Related Resources

MockUp : `link to mockup`
Coda doc: `link to coda`

sebjf:
This fixes #683.

#### Description

This PR makes a very small functional change, in that the default graph is cleared after `generateAndCommitSelectionTree` returns. 

However, bouncer often accesses the graphs by iterating over their members, and both the graphs and members are non-nullable. Therefore this PR also makes almost all accesses to the graphs go through a method that checks if the graph has been explicitly cleared, and throws an exception if this is the case, to make certain no accesses to default take place once it has been cleared.

The purpose of this change is to reduce the peak memory of bouncer by freeing the default graph once it is no longer used.

### DIFF
--- a/bouncer/src/repo/core/model/collection/repo_scene.cpp
+++ b/bouncer/src/repo/core/model/collection/repo_scene.cpp
@@ -97,6 +97,8 @@ RepoScene::RepoScene(
 {
 	graph.rootNode = nullptr;
 	stashGraph.rootNode = nullptr;
+	graph.cleared = false;
+	stashGraph.cleared = false;
 	//defaults to master branch
 	branch = repo::lib::RepoUUID(REPO_HISTORY_MASTER_BRANCH);
 }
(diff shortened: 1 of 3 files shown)


sebjf:
Dropping this for now as default graph is used in more places than expected and its not clear its worth the refactoring effort at the moment.

Issue #678, opened 2024-03-06
Large meshes are split across multiple submesh ids when they should be shared


sebjf:
### Description

When large meshes (65k vertices or more) are split, the splits result in new submesh ids, even when the geometry collector layer has not changed.

This results in mappings files like the one below.

In this file `bff2f76e-11b4-4c55-a659-7caa3121fc32` should have one entry with multiple usages, but instead there are multiple mappings for the one element.
This is not just a mappings issue - the geometry itself has multiple redundant ids.

This appears to cause problems with the rectangle select but the mechanism is not clear at this time. The number of mappings is different from `numberOfIDs` and `maxGeoCount` meaning these values cannot be used.

Currently, by chance these are not used in the current version of `AssetMapLoaded`, which is why this is not causing any issues - the files produced are perfectly consistent, just not as intended.

The likely candidate for the source of the bug is `MeshMapReorganiser::performSplitting()`.

The workaround is simply not to use `numberOfIDs` or `maxGeoCount`.

```
{
	"model": "b17270e2-dd46-452a-a7b9-aa5d0aaab1df",
	"supermeshes": [
		{
			"id": "91792cd0-dc9a-4814-aca0-f4d0546a5442",
			"data": {
				"numberOfIDs": 2,
				"maxGeoCount": 2,
				"mapping": [
					{
						"name": "be261ed1-4296-4de9-baf8-b0b756763e2a",
						"sharedID": "ffd297f4-3de8-4dd0-b246-1061712eb0e5",
						"min": [
							119.146484,
							0,
							-319.144531
						],
						"max": [
							319.146484,
							200,
							-119.144531
						],
						"usage": [
							"91792cd0-dc9a-4814-aca0-f4d0546a5442_0"
						]
					},
					{
						"name": "bff2f76e-11b4-4c55-a659-7caa3121fc32",
						"sharedID": "5c555269-6080-4c1a-9e54-e9669977322c",
						"min": [
							16.5898438,
							241.378906,
							-423.023438
						],
						"max": [
							421.699219,
							646.742188,
							-15.2734375
						],
						"usage": [
							"91792cd0-dc9a-4814-aca0-f4d0546a5442_1"
						]
					},
					{
						"name": "bff2f76e-11b4-4c55-a659-7caa3121fc32",
						"sharedID": "5c555269-6080-4c1a-9e54-e9669977322c",
						"min": [
							16.5898438,
							241.378906,
							-423.023438
						],
						"max": [
[...]

Issue #83, opened 2016-01-28
Update to MongoDB C++ Driver 3.0.0


jozefdobos:
https://github.com/mongodb/mongo-cxx-driver/releases

Once done, make sure to update also our own instructions. These at the moment point to 1.0.5 legacy: https://github.com/3drepo/3drepobouncer/wiki/Compiling-MongoDB-CXX-Driver

carmenfan:
NOTE: It looks like the c++11 driver uses CMake instead of scons so be sure to update travis script so it compiles for autotests

jozefdobos:
Ok, that would explain why scons was complaining so badly last night!

carmenfan:
In short, the entire API and the way it works is completely different to legacy CXX driver.

This issue is going to be parked until mongodb guys release a more developed version of the driver for 2 main reasons:

1. the API isn't set in stone yet, they want to get feedback from users and see how they should go about modifying it, if at all
2. It currently feels really green, there is hardly any documentation on the driver apart from the API. Only other thing out there is a brief tutorial which has very brief example of how to do simple things. 2 biggest concerns I've encountered so far is 

1.  the only way to connect and authenticate a user is to pass in a connection string in plain text, i.e. there is no way for us to store the digested password, and there is no way for us to keep a connection and change authentication (there seems to be some todos within the code on auth part within the database objection)
2. the connection procedure has no way of testing if you have been successfully authenticated or connected on the database until you try to do an operation that causes an exception to be thrown. For example I put in the right database address but wrong authentication, it "connects" fine, can list the databases (non existent) and not until i try to insert a document does it throw an exception. 

Granted everything can be worked around, the runCommand() function is available on the API so we could fill in the gaps ourselves. But it is probbaly not the best way to go as it is almost as if we're writing our own interface to the database if we rely on runCommand() and will be doing extra work when they decide to change their API

PR #685, opened 2024-05-29
Issue 682 - Decommission AssetBundleCreator 

### LINKED ISSUE #682: Decommission AssetBundle Support
### Description

Remove support for AssetBundles and make associated improvements as per this product ticket: https://github.com/3drepo/3D-Repo-Product-Team/issues/402

### Goals

- [x] Bouncer only builds RepoBundles and contains no references to the supporting infrastructure for AssetBundles
- [x] There are dedicated types for Meshes and Supermeshes

### Tasks

- [x] Remove 3drepobouncerwrapper and wrapper tests projects
- [x] Remove `AssetModelExport` class and its integration in scene manager
- [x] Add a new Supermesh type to bouncer that is in-memory only, to be used where `MeshNode` instances behave as Supermeshes (i.e. MultipartOptimizer and MeshMapReorganiser)
- [x] Move rest of code away from using builder objects and use just factory methods 
- [x] Move the submesh ids and mappings from `MeshNode` to the new type.
- [x] Make stash graph in memory only 
- [x] Make MultipartOptimizer use default graph UUIDs/references, since we aren't going to commit the stash graph so theres no need to avoid x contamination.
- [x] Address https://github.com/3drepo/3drepobouncer/issues/678
- [x] See if we can delete `insertBinaryFileToDatabase`
- [x] Move metadata into repoAssets
- [x] Rename RepoUnityAssets to RepoAssets
- [x] Add unit tests for Mesh Map Reorganiser
- [x] Update unit tests that use the stash graph (i.e. RepoBundles tests)

### Questions
1. Should we leave the `MISSING_BUNDLES` `UploadStatus` flag, given that it is hooked up outside bouncer?

### Related Resources
https://github.com/3drepo/3D-Repo-Product-Team/issues/402

sebjf:
This fixes #682 

Note: anything imported after this merge will require an updated viewer to see: https://github.com/3drepo/3drepounity/pull/534

#### Description

This PR removes support for `AssetBundleCreator`, and the written stash graph, since that is no longer needed with `AssetBundleCreator` gone.

This PR predominantly deletes decommissioned parts of the project, but also updates the RepoBundles export to write metadata that .io would have picked up from the stash graph, into the assets list document.

Additionally, the types have been cleaned up. `SupermeshNode` has been introduced as a subclass of `MeshNode`, and all supermesh specific properties are now members of this. The way these nodes are built throughout the code has been refactored to keep as much implementation detail in the factory methods as possible as well, instead of having classes modify the BSON objects themselves with the `BSONBuilder` & `clone` methods. 

As part of this, a new class `RepoMeshBSONBuilder` has been added for internal use by the BSON factory, for building `MeshNode`s and `SupermeshNode`s.

These changes are "light touch" however, and only done where trivial. The expectation is that big changes will occur in https://github.com/3drepo/3drepobouncer/issues/83, when we know exactly what needs to happen. 
For example, though the stash graph is no longer written to disk, its types (i.e. based on `RepoNodeSet`) are still the same, and accessed via `RepoScene` in the same way.

`MultipartOptimizer` has been refactored so that it no longer creates stash material nodes. This is because since the stash graph is no longer written, there is no longer a need for this indirection. When exporting both textures and materials are now accessed directly from the `Default` graph.

The `MeshMapReorganizer` has been given a set of unit tests, and an outstanding bug fixed. After fixing the bug, the RepoBundles exporter has updated how it builds the mappings, and a check in the RepoBundles unit tests has been re-enabled.

Finally, the `RepoBundleExport` unit tests have been revised so that they no longer rely on the stash graph.

Specifically, this PR:

1. Deletes the bouncer wrapper project & its unit tests
4. Deletes `AssetModelExport`
5. Deletes `insertBinaryFileToDatabase`
6. Updates the logic of `SceneManager` so that it no longer writes the stash graph or Unity AssetBundles.
7. Removes the ability to write the stash graph to the database, and deletes the `CommitStash` test
8. Turns `RepoUnityAssets` into `RepoAssets`, and updates the schema to include metadata that would previously have been extracted from the stash graph on-demand.
9. Introduces the `RepoBSONMeshBuilder` object for use by the BSON factory & reduces the direct use of BSON builder objects in the code in favour of factory methods
10. Moves some utilities created for the `MultipartOptimizer` unit tests into a common resource for also testing `MeshMapReorganizer`
11. Added unit tests for `MeshMapReorganizer`
12. Fixed https://github.com/3drepo/3drepobouncer/issues/678 & updated the RepoBundle exporter mappings logic accordingly
13. Adds `appendVector3d` method to write a Vector3D as an object instead of an array
14. Updates bouncer_worker to remove all references to the Unity Queue

Note, some of these changes are in https://github.com/3drepo/AssetGenerator/issues/3, which must be merged when this PR is.

This issue/PR also addresses: 

https://github.com/3drepo/3drepobouncer/issues/678
https://github.com/3drepo/3drepobouncer/issues/683


bouncer/src/repo/core/model/bson/repo_bson_builder.h
```diff
@@ -145,6 +145,14 @@ namespace repo {
 					mongo::BSONObjBuilder::append(label, date);
 				}
 
+				/**
+				* Appends a Vector but as an object, instead of an array.
+				*/
+				void appendVector3d(
```
carmenfan, line 151: would we maybe call it `appendVector3DObject`? 

just to be clear (as whoever is coding might miss the comment, esp if they're using intellisense!)

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
```diff
@@ -25,6 +25,203 @@
 
 using namespace repo::core::model;
 
+/**
+* Specialisation of the BSON Builder dedicated to building Meshes and
+* Supermeshes. This is for internal use by the RepoBSONFactory and is
+* intended to build immutable nodes via one of the makeMeshNode
+* overloads.
+*/
+class RepoMeshBSONBuilder : public RepoBSONBuilder
```
carmenfan, line 34: Is there any reason why it's hidden in here?

I understand it's not used elsewhere, but there's probably no harm in separating it out since it's a sub class of RepoBSONBuilder it would make more sense to live with the RepoBSONBuilder

or it's own file (probably prefer that)

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
```diff
@@ -25,6 +25,203 @@
...
+	* 2. leave mongo's bson, use our own/exteral library that doesn't have this limit and database handle this at the point of commit
+	* below uses option 1, but ideally we should be doing option 2.
+	*/
+
+	template<typename T>
+	void appendLargeArray(std::string name, std::vector<T> data)
+	{
+		uint64_t byteCount = data.size() * sizeof(data[0]);
+
+		if (byteCount + _byteSize >= REPO_BSON_MAX_BYTE_SIZE) //inclusion of this binary exceeds the maximum, store separately
```
carmenfan, line 62: if we're changing this anyway... 😆 

I believe we can just push it into binMapping regardless.. As we never store it as part of mongobson anymore (we always flush it out just before committing)

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
```diff
@@ -25,6 +25,203 @@
...
+		{
+			appendBinary(
+				name,
+				&data[0],
+				data.size() * sizeof(data[0]));
+			_byteSize += byteCount;
+		}
+	}
+
+	void appendDefaults(
```
carmenfan, line 81: from this part on, this is getting slightly weird because it's very much what the bson_factory was responsible for

Whereas bson_builder is more designed to be alight interface to building a bson, with understanding of 3drepo types.

for e.g appendDefaults is about appending the default information required for a RepoNode,  hence it lives in RepoBSONFactory as that has nothing to do with the concept of building a RepoBSON - it's a context of a specific type of repoBson (i.e. RepoNode)

So I'm starting to wonder why you feel like you need this new class - is this just to split the logic into multiple functions so you can pick and mix what it needs to do for mesh/supermesh?

I'm thinking it may be better instead to just have a helper function that takes a bsonbuilder and the data to build the common parts?

carmenfan: Or call this mesh node builder and remove the inheritance and have a repobsonbuilder as  an instance?

The blurring of roles between the factory and bson builder is what's confusing me a bit, not necessarily the concept of having a helper class here!

bouncer/src/repo/core/model/bson/repo_bson_factory.h
```diff
@@ -288,15 +265,22 @@ namespace repo {
...
+					const std::vector<repo::lib::RepoUUID>            &parents,
 					const int                                         &apiLevel = REPO_NODE_API_LEVEL_1);
 
+				static MeshNode makeMeshNode(
+					const std::vector<repo::lib::RepoVector3D>& vertices,
+					const std::vector<repo_face_t>& faces,
+					const std::vector<repo::lib::RepoVector3D>& normals,
+					const std::vector<std::vector<float>>& boundingBox,
+					const std::vector<std::vector<repo::lib::RepoVector2D>>& uvChannels,
+					const std::vector<repo_color4d_t>& colors);
```
carmenfan, line 282: looking at this it makes me wonder if we should still pretend that we support colors on a mesh when it doesn't get used anywhere.... 😆

bouncer/src/repo/core/model/bson/repo_node_mesh.cpp
```diff
@@ -173,7 +173,19 @@ RepoNode MeshNode::cloneAndApplyTransformation(
 	}
 }
 
-MeshNode MeshNode::cloneAndUpdateMeshMapping(
+SupermeshNode SupermeshNode::cloneAndUpdateIds(
```
carmenfan, line 176: I think this should live in RepoNode

as these fields are common for any repo nodes (and I feel like we already have something similar in that class... not 100% sure though)

bouncer/src/repo/core/model/bson/repo_node_mesh.h
```diff
@@ -230,16 +218,45 @@ namespace repo {
...
 				/**
-				* Given a mesh mapping, convert it into a bson object
-				* @param mapping the mapping to convert
-				* @return return a bson object containing the mapping
+				* Retrieve a vector of faces (serialised) from the bson object
+				*/
+				std::vector<uint32_t> getFacesSerialized() const;
+			};
+
+			class REPO_API_EXPORT SupermeshNode : public MeshNode
```
carmenfan, line 226: I think it'll be clearer to have this as a separate file (also the cpp)

bouncer/src/repo/core/model/bson/repo_bson_factory.h
carmenfan: we're also adding `{}` unncessarily everywhere to accomodate it...

I think we should either just remove it from the function, or supply a default value?

bouncer/src/repo/manipulator/modelutility/repo_scene_manager.cpp
```diff
@@ -352,35 +327,18 @@ bool SceneManager::generateStashGraph(
...
-			scene->updateRevisionStatus(handler, repo::core::model::RevisionNode::UploadStatus::GEN_REPO_STASH);
-		}
-
 		removeStashGraph(scene, handler);
+
 		repoInfo << "Generating stash graph...";
+
 		repo::manipulator::modeloptimizer::MultipartOptimizer mpOpt;
-		if (success = mpOpt.apply(scene))
+		if (mpOpt.apply(scene))
```
carmenfan, line 335: could we not do something like?
```suggestion
		if (!(success = mpOpt.apply(scene)))
```

Then we don't need the "then" branch

carmenfan (changes requested):

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
```diff
@@ -796,12 +803,33 @@ RepoUnityAssets RepoBSONFactory::makeRepoBundleAssets(
...
+		metadataBuilder.append(REPO_ASSETS_LABEL_NUMVERTICES, (unsigned int)meta.numVertices);
+		metadataBuilder.append(REPO_ASSETS_LABEL_NUMFACES, (unsigned int)meta.numFaces);
+		metadataBuilder.append(REPO_ASSETS_LABEL_NUMUVCHANNELS, (unsigned int)meta.numUVChannels);
+		metadataBuilder.append(REPO_ASSETS_LABEL_PRIMITIVE, (unsigned int)meta.primitive);
+		metadataBuilder.appendVector3d(REPO_ASSETS_LABEL_MIN, meta.min);
+		metadataBuilder.appendVector3d(REPO_ASSETS_LABEL_MIN, meta.max);
+		metadataNodes.push_back(metadataBuilder.obj());
+	}
+
+	if (metadataNodes.size())
```
carmenfan, line 829: @sebjf  started reading https://github.com/3drepo/3drepounity/pull/534/files and this thought came up...

Since this is stored as a bson object in mongo, are we certain that by adding the metadata nodes here, we dont' risk pushing the size of the bson beyond 16MB (which is the limit per document on the database)?

carmenfan:
@sebjf  another thing that's missing here (not sure if you're planning to make it a separate ticket)

We need to change the bouncerworker so it doens't plonk the task into unityq after completing the work

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
sebjf: Done!

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
sebjf: Hi Carmen, ah I understand...

I have turned `RepoMeshBSONBuilder` into `RepoBSONBinMappingBuilder` and moved it to its own file. I have made it responsible for managing the binary mappings internally, and kept the `appendLargeArray` method, but the other methods have been made private members of `RepoBSONFactory`. 
I have also made sure there is only one set of `appendDefaults` methods, common to the factory, that the mesh building methods use.

bouncer/src/repo/core/model/bson/repo_node_mesh.h
sebjf: Done!

bouncer/src/repo/core/model/bson/repo_bson_factory.h
sebjf: Colour support is finally removed 😆

bouncer/src/repo/manipulator/modelutility/repo_scene_manager.cpp
sebjf: Done!

bouncer/src/repo/core/model/bson/repo_node_mesh.cpp
sebjf: Hi @carmenfan, the problem with putting this in `RepoNode` is that that will return the base class, whereas the caller expects a `SupermeshNode`. However I've thought again and a neater way is just to not have it and make `buildMesh` take the Ids, which is closer to the goal of treating meshes as immutable once built anyway!

bouncer/src/repo/core/model/bson/repo_bson_factory.cpp
sebjf: Hi @carmenfan, good point! I have worked this out...

Using the objsize() method of the assets list document returned by `makeRepoBundleAssets` (which I assume is an appropriate measure because that's what we used to do the bin mapping before!), the document size increases just sub-linearly with the number of entries, starting at 263 bytes per entry.

So, we could fit 63,791 assets into a document before we run out of space.

If each supermesh had the maximum number of submeshes (5000), that would mean a container could have 318,958,479 elements before we hit the limit on this document.

I have added a unit test that is pretty much at this limit, so if the schema changes we'll get a warning to consider this again.

sebjf:
> @sebjf another thing that's missing here (not sure if you're planning to make it a separate ticket)
> 
> We need to change the bouncerworker so it doens't plonk the task into unityq after completing the work

Hi @carmenfan, this one fell through the cracks as I thought it was devops' scope, but I can make the change here!

I am not familiar with bouncer worker, but if I understand right...

```
		callback(JSON.stringify(returnMessage), config.rabbitmq.unity_queue);
		callback(JSON.stringify({
			status: UNITY_QUEUED,
			database,
			project: model,
		}));
```

The first line puts the model in the unity queue where, it's picked up by `unityQueueHandler` `onMessageReceived`, and the second line updates the status for the frontend?
So what we need now is to just have one invocation of `callback` passing `returnMessage` without the queue parameter?

sebjf:
@carmenfan, ready for another look!

bouncer/src/repo/core/model/bson/repo_bson_factory.h
```diff
@@ -263,33 +277,42 @@ namespace repo {
...
-					const std::vector<repo_color4d_t>& colors);
+					const std::vector<std::vector<repo::lib::RepoVector2D>>& uvChannels);
+
+				static MeshNode makeMeshNode(
+					const std::vector<repo::lib::RepoVector3D>& vertices,
+					const std::vector<repo_face_t>& faces,
+					const std::vector<repo::lib::RepoVector3D>& normals,
+					const std::vector<std::vector<float>>& boundingBox)
+				{
+					return makeMeshNode(vertices, faces, normals, boundingBox,
```
carmenfan, line 301: A bit confused why we have 3 different function declarations for makeMeshNodes

Can't we just have 1 with default values declared for optional parameters?

bouncer/src/repo/core/model/bson/repo_bson_factory.h
```diff
@@ -322,13 +343,28 @@ namespace repo {
...
 					const std::vector<std::vector<float>>& boundingBox,
 					const std::vector<std::vector<repo::lib::RepoVector2D>>& uvChannels,
-					const std::vector<repo_color4d_t>& colors,
 					const std::vector<float> mappingIds,
 					const std::vector<repo_mesh_mapping_t>& mappings,
 					const repo::lib::RepoUUID& id,
 					const repo::lib::RepoUUID& sharedId
 				);
 
+				static SupermeshNode makeSupermeshNode(
```
carmenfan, line 352: if we rejig the param orders then we can make mappingId optional and have one function declaration?

carmenfan (changes requested):
looks fine, just 2 small organising related questions

bouncer/src/repo/core/model/bson/repo_bson_factory.h
sebjf: Done!

bouncer/src/repo/core/model/bson/repo_bson_factory.h
sebjf: Done!

carmenfan (approved):

PR #696, opened 2024-08-30
Issue 693 - Process PDFs

### LINKED ISSUE #693: Process PDFs to SVG
### Description
Initially it was planned to be done on 3drepo.io, however it turns out there is a limited amount of PDF to SVG capability on existing node js libraries, and also a big PDF will take a significant amount of RAM and processing power.

So it has been decided to put this in the bouncer queue as well. The actual processing method is still TBC... We may do this on bouncer-worker calling inkscape (as this is the easiest approach), there are more integrated possibilities with paid libraries like [PDFTron](https://docs.apryse.com/documentation/samples/cpp/ConvertTest/), or using something like [poppler](https://github.com/innodatalabs/poppler?tab=readme-ov-file) (GPL-v2)

### Goals
- [ ] As a user I want to be able to convert the first page of my PDF into a SVG

### Tasks
- [x] Process pdf
  - [x] Determine if drawing import is PDF
  - [x] Get inkscape to produce a svg
    - [x] Ensure processInfo is a union of both process instead of overwriting
    - [x] Error doesn't exit in failure
  - [x] get 3drepobouncer to add the svg into the system
- [x] Unify mimetype naming (we're using `mime` in `drawings.history.ref` but it's `mimeType` everywhere else
- [x] Clean ups
  - [x] Why do we need revId in the import params.json when it's basically the corId?
    - we need it for bouncerclient to fetch the file, but bouncer_worker doesn't need to read it.
  - [x] Store the svg pre-zipped?
  - [x] Check inkscape can be found at launch of bouncer-worker

carmenfan:
This fixes #693

#### Description
NOTE: This requires https://github.com/3drepo/3drepo.io/issues/5108 

- Added support to process import drawing from pdf format
  - it will first call inkscape to generate the svg and dump it in the sharedDir
  - it will then call the bouncer command to import drawing with the file path to svg
    - bouncer has been altered so if a image path has been provided, we will bypass the import process and just save the image file
- `mime` has been renamed to `mimeType` to match the usage in 3drepo.io
- process monitor will now accumuate run time and take the maximum memory usage if more than 1 process is called for that paritcular rId (e.g. inkscape and bouncerclient)
- SVGs are now gzipped before storing

#### Test cases
- importing dwg drawing should work as before
- importing pdf should now work


sebjf (commented):
Looks good!

sebjf (approved):

PR #704, opened 2024-09-24
Issue #698 ODA Upgrade to 2025.8

### LINKED ISSUE #698: ODA Upgrade
### Description

The bouncer is currently running on Teigha_2024.8, which prevents files from the 2025 versions of the autodesk tools to be imported. This necessitates an update of the ODA version. It is expected that some things might break in the process that will need to be fixed.  

### Goals

- [x] The bouncer should still accept previous file versions. 
- [x] The bouncer should accept 2025 file versions.

### Tasks

- [x] Add new Teigha/ODA version as dependency to the build system, replacing the previous version.
- [x] Fix compile issues that are most likely arise with the new version.
- [x] Fix runtime issues that are likely to arise with the new version.
- [x] Test with old formats for regression issues.
- [x] Test with new formats to ensure compatibility.

### Related Resources

Product Ticket: [#384](https://github.com/3drepo/3D-Repo-Product-Team/issues/384)

FJThiel:
This fixes #698 

#### Description

Update the ODA Libraries to 2025.8 to allow import of new version of the file formats.
Main changes were:
- Creation of the Gi context has changed slightly with a dedicated method being added instead of the generic createObject()
- Retrieving of group labels has changed slightly with a dedicated method being added 
- Addition of new unit tests for 2025 RVT and NWDs. Additionally, the unit tests have been tweaked to cover the composite-object behaviour.
- The Revit materials handling has been updated to take parameters from the Shading members, and to override the colour when textures are used (and found) - this more closely matches the Revit viewport. This is not a regression issue but became very noticeable during testing.
- A new class RepoGlobalManager, is introduced that is used to destroy objects in the global scope, without the caller (which is usually a different project entirely) knowing about those objects themselves. This is required to clean up ODA since ODA have 'fixed' something in 2025, that now means the `ExSystemServices` class breaks when used more than once in the lifetime of a single process. A new class, `RepoSystemServices` has been introduced instead which connects to `RepoGlobalManager` in order to clean up ODA on-demand. (Note, the pattern of using the destructor of a global instance in the module doesn't work for ODA: it relies on cleanup happening before the end of main, probably because it itself is using a similar pattern underneath...)

#### Test cases

All existing regression tests are expected to pass.
Note: the test RepoBundles.SampleHouse will fail until[ this PR from Asset Generator](https://github.com/3drepo/AssetGenerator/pull/6) is merged as well.


sebjf:
Have looked through this and all appears to be working OK,

The biggest change is that the materials in Revit have been "improved" in the sense they more closely match the viewport in Revit itself:

image

In the image above the new behaviour is on the left and staging is on the right.

In the new bouncer, the kitchen units have a colour whereas before they were black making their shape difficult to discern. The fireplace has a colour, and the texture on the couch is visible. The opacity of the glass is also more accurate to the Revit viewport.

sebjf (dismissed):
Good to merge once it passes travis.

sebjf:
Hi @carmenfan,

This is now passing. I have approved Felix's changes, though I have made enough myself that you should probably take a look also! I have updated the description with these additional changes.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/repo_oda_system_services.cpp
carmenfan, line 1: since the class is `RepoSystemServices` i think it's better to rename this file to `repo_system_services`

The nature of this being inside the odaHelper folder should be obvious enough that it's ODA related.

bouncer/src/repo/repo_global_manager.h
carmenfan, line 1: this should live in the `lib` folder as it's a helper class

bouncer/src/repo/repo_global_manager.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "repo_global_manager.h"
+#include "repo/lib/repo_exception.h"
+#include <stack>
+
+static bool initialised = false;
+static std::stack<std::unique_ptr<repo::RepoGlobalManager::Destructor>> destructors;
+
+REPO_API_EXPORT repo::RepoGlobalManager::RepoGlobalManager()
+{
+	if (initialised) {
```
carmenfan, line 27: would it be better to just use a singleton pattern? 

so it doesn't matter if more than one place instantiates it, and the whole thing doesn't need to be a static class (which feels a bit weird)

And the class itself doesn't need to be instantiated if nothing needs to be handled and deconstructed at the end of the application.

Also then we don't need external applications linking to the library to be aware this needs to be instantiated, which should hopefully make it less prone to error

carmenfan (changes requested):

bouncer/src/repo/repo_global_manager.cpp
sebjf: @carmenfan 

> would it be better to just use a singleton pattern?

I wish, but unfortunately not!

I tried that originally, but I believe that ODA also uses the pattern internally, because when we rely on the destructor of a global variable the clean up results in an error, which being outside of main we can't catch, leading to an unexpected exit code which confuses bouncer worker.

I don't think its a good idea to try and manipulate the compiler into instantiating our global first, because that isn't a supported feature and could break at any time.

Instead of a scoped variable in main, we could just have a global `cleanup` function, but the trouble is that then needs to be called wherever there is a return statement. The way it is now at least we don't need to track that - we just instantiate the object and forget about it.

Though, a third option is simply not to do anything...

Thinking about it, it could be better just to let the process finish and have the operating system clean up anything leaked related to ODA.

The only method we are trying to call is `odCleanUpStaticData`, and this should only be called right before main returns anyway, so we're putting in all this effort to follow best practice, but there's no practical behavioural difference. (And ODA is clearly doing something funky anyway, otherwise there wouldn't be any problem initialising multiple times.)

carmenfan: Just to know if i understood this correctly...

so the problem is ODA objects might've been instianted first because at the point of us calling the singleton if we only call it when we need it?

What if we initialised the singleton when we initialise the repo controller? I think the main issue i have here (besides everything being static 😆 ) is that we rely on the bouncerclient to know to call it, which means we rely on the third party callee outside of the library to know this is necessary - which makes it more error prone and messy


> Thinking about it, it could be better just to let the process finish and have the operating system clean up anything leaked by ODA.

That's fine, since the nature of how bouncer is likely to be that it exits as soon as it's done with the processing. But I thought the reason you had to do this was to do with tests and ensuring some functions are only ever called once?

sebjf: > so the problem is ODA objects might've been instianted first because at the point of us calling the singleton if we only call it when we need it?

Yes, but specifically only objects related to licensing. Most ODA resources behave as we'd expect, and can be initialised and cleaned up with dedicated methods many times. However, for their licensing, something different happens where there some resources are allocated "somewhere" and they can only be cleaned up once, and never used again.

I would've hoped that this would only be after calling `odActivate`, but it seems not to be the case - my best guess is that they have a global variable that is initialised at the point of declaration, so we have no idea what the compiler is doing regarding the ordering of that vs. ours.

> What if we initialised the singleton when we initialise the repo controller

This would work in the sense it would be cleaned up, but then as soon as the first RepoController instance is destructed, we would not be able to use ODA again, which would be unexpected and break processes that instantiate multiple controllers over time, like the unit tests.

> I thought the reason you had to do this was to do with tests and ensuring some functions are only ever called once?

Kind of.

A number of our unit tests create and tear down an ODA system (implemented in the `OdRxSystemServices` class). Initialising multiple ODA systems over time is supported. What is not supported is activating and deactivating (cleaning up the activation static resources with `odCleanUpStaticData`).

ODA's sample `OdRxSystemServices` subclass however cleans up the activation info, meaning it can only be used once. 

So, what we do *have* to do is stop `odCleanUpStaticData` being called. This is straightforward, by making our own `OdRxSystemServices` subclass.

However, doing this gives us the responsibility of deciding when `odCleanUpStaticData` should be called, since the ODA class doesn't do it anymore.

Although, we don't *need* to call it, it's just good practice (e.g. if we ran bouncer through valgrind or something it would complain).

> That's fine, since the nature of how bouncer is likely to be that it exits as soon as it's done with the processing. 

Great, I agree in this case its better just to let it leak, given how messy the alternative is, and we are only in this situation because ODA's doing something odd with the licensing code. I will confirm that bouncer still terminates properly without it and remove the helper.

sebjf:
This is ready for another look @carmenfan!

carmenfan (approved):

PR #710, opened 2024-11-04
ISSUE 703 - Mongo Driver Upgrade Part 2 - Refactor

### LINKED ISSUE #703: Mongo Driver Upgrade - Mongo type decoupling
### Description

Upgrade the mongo driver; [product ticket](https://github.com/3drepo/3D-Repo-Product-Team/issues/561) with further details.

This issue is for the decoupling of the mongo dependencies, which is also described by this ticket: https://github.com/3drepo/3D-Repo-Product-Team/issues/560

This branch relies on https://github.com/3drepo/3drepobouncer/issues/697

### Goals

- [ ] RepoBSON no longer inherits from BSONObj publicly
- [ ] RepoBSON and RepoBSONBuilder are isolated to the minimum number of modules possible (Nodes and Bsons)
- [ ] There are no mongo references outside of MongoDatabaseHandler

### Tasks

- [x] Refactor DrawingRevisionNode
- [x] Refactor MaterialNode
- [x] Refactor MeshNode
- [x] Refactor MetadataNode
- [x] Refactor ModelRevisionNode
- [x] Refactor ReferenceNode
- [x] Refactor RevisionNode
- [x] Refactor SupermeshNode
- [x] Refactor TextureNode
- [x] Refactor TransformationNode
- [x] Update RepoNode tests
- [x] Update DrawingRevisionNode tests
- [x] Update MaterialNode tests
- [x] Update MeshNode tests
- [x] Update MetadataNode tests
- [x] Update ModelRevisionNode tests
- [x] Update model revision handling to make file prefix not implicit
- [x] Update ReferenceNode tests
- [x] Update TextureNode tests
- [x] Update TransformationNode tests
- [x] Update RepoTask
- [x] Create RepoTask unit tests
- [x] Create RepoSequence unit tests
- [x] Update RepoCalibration
- [x] Make RepoProjectSettings unit tests
- [x] Update RepoProjectSettings
- [x] Go through header files to clean up, make classes forward declared where possible
- [x] Remove filename support from bin mappings
- [x] Refactor binary file handling so they come from inside the bin mappings only and remove the shrink option during writing the RepoScene
- [x] Update RepoBSON unit tests
- [x] Add BSON Element unit tests
- [x] Update RepoBSONBuilder unit tests
- [x] Update MongoDatabaseHandler unit tests
- [x] Refactor all the cloneAndAddParent calls
- [x] Run linter
- [x] Extend Trans Optimiser Tests
- [x] Run full suite of unit tests
- [x] Refactor the metadata stuff in files handler to use variants
- [x] Update factory methods to use the methods/constructors of the mutable types
- [x] Get it to build on linux
- [ ] Fix all tests on Travis

No longer doing as part of this: 

- [ ] Replace usage of repo_material_t with MaterialNode
- [ ] Make a criteria type to hand to database handler so other types can stop relying on the builder
- [ ] Refactor to forward declare all mongo types in mongo database handler header

### Manual Tests

- [x] Drawing DWG/Autocalibration
- [x] Drawing PDF/Manual Calibration
- [x] ODA Import (Revit House, MCFC)
- [x] Revit Systems Import
- [x] NWD Federation Import
- [x] Revisions & Diff (Clinic A)
- [x] IFC Import (Impraria)
- [x] SPM Import (Clinic S Landscape, Displacement Test)
- [x] Hidden Geometry (Clinic S w/ Landscape)
- [x] FBX Import (Mixed Resolution, Mixed Textured)
- [x] OBJ Import (Multiresolution, Cubes)
- [x] Lines only OBJ Import (Gamma)
- [x] DWG as Model Import (Revit House, Coloured Boxes 3D, Car Park)
- [x] DGN as Model Import
- [x] Sequences and Sequences Color Overrides (Clinic S Landscape, Displacement Test)
- [x] Create, Modify Federation (Car Park)
- [x] Modify Legacy Federation (Empty Containers, Groups)
- [x] genStash on Legacy Import

sebjf:
This fixes #703

#### Description

This PR is to encapsulate the mongo types, keeping them behind a well-scoped & unit-tested API behind which the mongo API & driver can be changed with minimum risk.

*RepoNode vs RepoBSON*

`RepoNode` becomes the new base class for database types in `repo::core::model`. `RepoBSON` represents an actual database document. `RepoNode` is used in `RepoController` and `RepoManipulator` to set up the scene, drawing etc, and then when finalised is converted into a `RepoBSON` through implicit conversion operators for interaction with the database handler. `RepoBSON` is the type used (exclusively) by the database handler API.

*Inheritance and Conversion*

`RepoNode` does not inherit from `RepoBSON`, but can be constructed from & implicitly cast to a `RepoBSON` (deserialisation & serialisation).

For example, to read an implicitly typed `RepoNod`e from a collection:

```
auto revision = DrawingRevisionNode(handler->findOneByUniqueID(teamspace, REPO_COLLECTION_DRAWINGS, revision));
```	

And to write it back:

```
handler->upsertDocument(teamspace, REPO_COLLECTION_DRAWINGS, revision, false, error);
```

`RepoBSON` still inherits from `mongo::BSON`, but that inheritance is now private. `RepoNode` cpp modules reference `RepoBSON` and `RepoBSONBuilder`, which themselves hide any mongo APIs. `RepoBSONFactory` no longer directly references `RepoBSONBuilder`, but must go through the public APIs of types that can cast to `RepoBSON`.

*Types*

The `RepoNode` subclass is set in stone the first time a `RepoBSON` is deserialised (i.e. to whatever constructor was called), so must be converted to the most refined subclass in the first instance (so it is not possible to do: `auto mesh = (RepoMesh)(RepoNode(RepoBSON)`, for example).

The `RepoBSON` member `getType()` returns the dynamic type stored in the database for a document. The `RepoNode` `getType()` member returns a compile-time constant [1].

RepoNodes can however be assembled into heterogenous sets (e.g. { &MeshNode &MeshNode &MaterialNode }) [4]. RepoNodes also interoperate with the RTTI. For example, converting from a superclass set is done like so:

```
auto nodes = std::vector<RepoNode&>({ meshNode meshNode materialNode });
auto meshNode = dynamic_cast<MeshNode&>(nodes[0]);
```

The `dynamic_cast` provides runtime type checking. Above it will throw a `bad_cast` exception if `nodes[0]` is not a `MeshNode`, or return null if used with pointers.

*Serialisation & Deserisalistion*

`RepoBSON` acts as the serialisation/deserialisation interface. It inherits from BSONObj privately, so only it is allowed to interact with the mongo API [2].

*RepoBSON*

`RepoBSON` now supports directly parsing types such as `RepoMatrix` and 3D bounds, which previously were parsed by the nodes that used them, making the storage of these types an implementation detail & ensuring consistency where used between nodes.

`swap` is now private.

`getListStringPairField` has been removed, along with `getEmbeddedDouble` and `hasEmbeddedField`, `getCurrentTimestamp`, `isValid` and `couldBeArray`. 

`RepoBSON` method definitions have been moved into the module, even for simple ones, to reduce exposure of the mongo types in the header. repo_bson.h now only requires the mongo BSON Obj header for its private inheritance.

New exceptions have been introduced: `RepoFieldTypeException` and `RepoFieldNotFoundException`, for use when parsing RepoBSONs. `RepoBSON` now throws if it cannot successfully read a field (the same behaviour as the 'capitalised' mongo methods, String(), Int(), Double() etc.) It is now up to the caller to check for fields beforehand or catch the exception, and then determine if these are schema errors & if so if they are fatal or not.

`RepoBSONElement` has also had most of its members moved into its module. The `date()` member has been removed and it now returns either `time_t` or `tm` directly. Similarly, `toMongoElement()` has been removed (the `mongo::BSONElement::getObjectID()` method is used to get the `_id` field inside the database handler instead, allowing the mongo type to be hidden from the outside).

*Bin Mappings*

All large arrays are now handled by `RepoBSON` exclusively. The `BinDataGeneral` type of mongo is no longer accessible & `RepoBSON` no longer allows storing or reading `BinDataGeneral` fields directly (i.e. via `getBinary` or `getBinaryAsVector`).

RepoNodes no longer hold their own `bigFiles` member. Instead, each member is serialised/deserialised from the appropriate store explicitly; for example, vertex attribute arrays use `RepoBSONBuilder::appendLargeArray` and `RepoBSON::getBinaryFieldAsVector`, both of which operate on RepoBSON's `bigFiles` array, while `RepoUUID` is stored as traditional mongo binary field type. The details of both are hidden inside `RepoBSON`.

`RepoBSONBinBuilder` has been merged with `RepoBSONBuilder`, so there is now only `RepoBSONBuilder`.

The `filename` component of `bigFiles` has been removed; blobs are now referenced by field name only. `_extRef` entries can still be read, but the `filename` will be ignored.

Oversized files (`_extRef`) support has been decommissioned further: the `getFilesList` & `hasLegacyFileReference` methods have been removed, and it is only possible to initialise one of these legacy documents through `createRepoBSON` (where it becomes a new mapping without the filename). This is in response to the 5.4 migration which should have moved all `_extRefs` to `_blobRefs` for `.scene` and `.stash.3drepo`; if we are sure there are no `_extRefs` outside these collections we could remove this snippet too - though at least one dumped collection used by the unit tests still uses `_extRef`.

`cloneAndShrink` is no longer supported, as all types should now send binary data to the right place during serialisation.

*Mutable Types*

RepoNodes are now fully mutable. 

In most places the `cloneAndAdd` pattern was attempting to modify a node in place, for example,

```
*mesh = mesh->cloneAndAddFields()
```

or 

```
auto updated = mesh->cloneAndAddFields()
mesh.swap(updated)
```

In these cases, the clone method has been replaced with an alternative that mutates the `RepoNode`.

Where the purpose of the clone was ambiguous, it has been left as a clone, but using the copy-constructor.

*Node Tests*

Based on the above, the pattern of the node unit tests typically look like the following:

1. Test the constructor to ensure default values are initialised correctly
2. Test that the public API modifies the type correctly
3. Test the Serialisation to RepoBSON (and so define the expected schema), ensuring that changing each field one at a time affects the schema correctly
4. Test the Deserialisation from RepoBSON
5. Test that the factory method (which may have a different set of defaults to the constructor) works
5. Test the copy constructor and sEqual

*Repo Documents*

`RepoTask`, `RepoSequence`, `RepoAssets`, `RepoProjectSettings` and `RepoCalibration` represent write-once database objects. For the purposes of this ticket I'll call them 'repo documents', as compared to RepoNodes. They follow a similar pattern to `RepoNode`, but with a couple major differences:

1. RepoNodes are entirely mutable by default, whereas repo documents are initialised through their constructors, and are modified as the exception.
2. RepoNodes are intended to serialise and deserialise, whereas repo documents are only intended to serialise.

Like RepoNodes, serialisation is effected through an implicit conversion operator to `RepoBSON`. The types no longer inherit from `RepoBSON` however. Immutability is now a consequence of what accessors are made available. Typically, repo documents do not provide any accessors, though there are exceptions.

`RepoSequence` has a method to check its size. Internally this converts to `RepoBSON`. We may to figure out an appropriate frame limit instead if desired, though leaving it like it is means the test will change depending on the driver used, which might be a good thing.

For `RepoProjectSettings`, unused accessors have been removed.

*RepoRef*

`RepoRef` is special in a couple of ways. It is not a `RepoNode`, but unlike repo documents it is expected to be read back, but only partially, and possibly from a document created from another source. Additionally, RepoRefs can be indexed by LUUIDs or Strings [3].

`RepoRef` has undergone a number of changes to clean it up.

`RepoRef` itself is now a superclass with a templated subclass for specifying the Id type. It is expected that instances of `RepoRef` are always one of the subclasses, created either via `RepoBSONFactory`, or the database handler. The Id can only be accessed via this subclass, ensuring type safety. It is possible to infer this and perform a dynamic cast from a `RepoRef` using RTTI, if desired.

The metadata functionality of `RepoRef` is now supported via RepoVariants, instead of a BSON object. `RepoRef` can only write metadata, not read it back.

The file handling APIs have been updated to use an unordered map of RepoVariants for metadata throughout as well, with independent type aliases in both `FileManager` and `AbstractDatabaseHandler`, to avoid introducing dependencies solely for that one alias.

Name is a read-only member.

*Other changes*

The `modifyNode` logic has been removed from `RepoScene`, as we no longer have any operations that read a scene revision and modify it in place. This logic relied on the swap ability of the mongo types, which we don't use anymore. If we need to version existing scenes in the future, then we need to introduce this another way. For example, a stage that compares the current scene state with the serialised version based on the unique id and duplicates and updates conflicts on demand.

`RepoMatrix` has been given some minor additional functionality.

The test utilities have been extended; tests have been updated where appropriate to use the [gmock matchers](https://google.github.io/googletest/reference/matchers.html) functionality for clarity.

The (original uploaded) file handling in `RepoScene` has been updated, so `RefNode` name is now explicitly created and assigned to the revision in `RepoScene`, rather than have `RepoScene` rely on the `RepoBSONBuilder` implementation detail.

Time handling has been updated so it operates with `tm` where possible, or `time_t`. Any use of `time_t` now follows the convention of the [ctime](https://cplusplus.com/reference/ctime/) methods - that is, `time_t` is the number of *seconds* since the unix epoch. (This is only where variables are stored explicitly as time; tasks or sequences that use longs everywhere are unaffected.)

The `ReferenceNode` API terminology has been updated, as some of these method names had become very confusing over time.

The `RepoBSONFactory` API has remained the same. The implementation creates RepoNodes or other types using only their public APIs, and no longer references `RepoBSONBuilder`. The factory methods may have different defaults to the constructors. (Generally speaking, type constructors do what 'C++' would expect, and initialise to type defaults, whereas `RepoBSONFactory` does what 'we' expect, and implements conventions such as some types automatically get a SharedId and some don't, etc.)

The `RepoMatrix` row-major storage order is now an implementation detail. Callers must use `transpose()`, rather than reading it in a different order, to get the transpose.

`RepoUUID::fromBSONElement` has been removed and the conversion now takes place in `RepoBSON`.

`binaryStorageMetadata` is no longer written by `RepoScene` - `RepoRef` documents should not rely on this to be resolved correctly, and the whole BSON-binary pipeline should no longer be used now binaries are handled explicitly in the serialisation methods.

`RepoVariant` now supports RepoUUIDs.

The `RepoUUID` const to non-const assignment operator has been removed, as this was creating a copy internally, and so would likely result in logic errors where the compiler silently removed the const modifier.

The header files for all the types touched have been updated to follow best practices. Reliance on transitive includes have been reduced. `RepoBSON` & `RepoBSONBuilder` are now forward declared where possible and only included where they are explicitly required in *modules*. repo_bson.h, repo_bson_builder.h and repo_database_handler_mongo.h are not included unless those actual types are needed by the module, and they are never included in other header files.

A new type, `RepoTeamspace`, has been introduced so `RepoSceneManager` no longer parses `RepoBSON` objects itself when looking for add-ons.

A number of legacy FIXME's & TODOs have been addressed.

`RepoManipulator` now takes the database user directly, instead of credentials document object, and no longer takes credentials in `getAllFromCollectionTailable` (this parameter was unused).

For all the unit tests touched, any Mongo Macros (e.g. BSON stream operators) have been replaced with `RepoBSONBuilder`, so all such operations go through our types exclusively (except where we are explicitly testing those types).

An existing bug in the RepoBundles asset list construction has been fixed.

*Future Work/what has NOT Been done*

[1] It should now be possible to use RTTI instead of the `type()` member on RepoNodes, but this change has not been made yet.

[2] BSON macros are still used to build queries, predominantly in the repo_scene.cpp module, but elsewhere too. These are not present in any headers. In the next stage though, these could/should be replaced with dedicated objects that encapsulate the query.

It might be a good idea to flag some documents as update-only; this can be used to check if they are being upserted with the overwrite flag set, in order to catch logic errors.

[3] `FileManger::RepoRefT` could be changed to use a variant to decide the index type, rather than a template. This is because it's too easy to assign a `RepoRefT` to a `RepoRef`, and so lose the `_id` field. The code would also be cleaner due to the number of explicit specialisations required to support it the way it is. On the other hand though, type-safety is enforced at compile time whereas using a variant errors would manifest until runtime.

The tests now often utilise a pattern whereby the random number generator is reset in order to get procedural functions to output the same values. We could consider wrapping `std::srand` and `rand()` in a dedicated class to better implement this.

[4] At least one bug has been found because the `RepoScene` constructor takes a `RepoNodeSe`t, instead of strongly-typed arrays. We could change all instances of `RepoNodeSet` to be strongly typed so this cannot happen again.

`RepoNode` could issue a warning or error if one of/the parents is set to the default UUID (after double checking if this is ever a supported case - I don't think so?).

Bounding boxes use a mix of types across bouncer (vectors of std vectors, vectors of doubles, vectors of RepoVectors of different storage type...). We should replace these all with a dedicated type & only make the change on serialisation. Apart from being neater, the new type could be better unit tested to reduce the chance of things like initialising the max with FLT_MIN instead of -FLT_MAX, and other common mistakes.

#### Test cases

In addition to the unit tests, a number of manual tests were made, including:

* ODA import with RVT & NWD files, textured and non-textured, with revisioning
* DWG as Drawings with auto-calibration and revisioning
* PDF as Drawings with manual-calibration and revisioning
* DWG and DGNs as Models
* Revit Systems support
* Navis (NWD) Federations (of RVT & IFC) support
* Assimp importer with OBJ and FBX files, intra-file mix of textured, non-textured, triangles & lines, and multiple resolutions, with revisioning
* IFC import with revisioning
* SPM import with sequences & tasks, colour overrides in sequences, hidden geometry and revisioning
* Diff and Clash
* Create & Modify Federations
* Modify Federations created with a legacy bouncer
* `genStash` on models imported with a legacy bouncer

The above found 6 bugs not captured by the unit tests:

* 3 were due to mistakes in the conventions introduced during the refactor. They are now captured by updated unit tests.
* 1 was due to an incorrect type assignment introduced during the refactor, and could be prevented in future by using strongly-typed containers.
* 1 was a legacy bug that has been fixed.
* 1 was in .io and has had a product ticket raised.

Finally, this PR has a [counterpart](https://github.com/3drepo/AssetGenerator/pull/8) on AssetGenerator that should be merged & the submodule here updated before the issue is completed.


bouncer/src/repo/core/model/bson/repo_node_metadata.cpp
```diff
@@ -29,61 +29,114 @@ RepoNode()
...
+		auto arr = bson.getObjectField(REPO_NODE_LABEL_METADATA);
+		auto names = arr.getFieldNames();
+		for (auto& n : names) {
+			auto field = arr.getObjectField(n);
+			auto key = field.getStringField(REPO_NODE_LABEL_META_KEY);
+			if (field.hasField(REPO_NODE_LABEL_META_VALUE))
+			{
+				metadataMap[key] = field.getField(REPO_NODE_LABEL_META_VALUE).repoVariant();
+			}
+			else
```
carmenfan, line 54: I don't think it's ever possible that we have a key field but no value?

Maybe it's better to throw an error if we find ourselves in this situation

bouncer/src/repo/core/model/bson/repo_node_model_revision.cpp
```diff
@@ -20,93 +20,89 @@
...
 }
 
-std::vector<std::string> ModelRevisionNode::getOrgFiles() const
+bool ModelRevisionNode::sEqual(const RepoNode& other) const
 {
-	std::vector<std::string> fileList;
-	if (hasField(REPO_NODE_REVISION_LABEL_REF_FILE))
-	{
-		RepoBSON arraybson = getObjectField(REPO_NODE_REVISION_LABEL_REF_FILE);
+	auto otherMesh = dynamic_cast<const ModelRevisionNode*>(&other);
```
carmenfan, line 95: should rename the variable as it's not a mesh 🙈

bouncer/src/repo/core/model/bson/repo_node_model_revision.cpp
```diff
@@ -20,93 +20,89 @@
...
-		std::set<std::string> fields = arraybson.getFieldNames();
-
-		for (const auto &field : fields)
-		{
-			fileList.push_back(arraybson.getStringField(field));
-		}
+	bool success = false;
+	if (otherMesh != nullptr)
+	{
+		success = true;
```
carmenfan, line 100: do we need to set it to true first? wouldn't it be the same as getting L101 to do an assignment instead of &=?

carmenfan (changes requested):
It's struggling to compile the test for me:
image

Quick google seems to think it's to do with requiring an env-var https://github.com/facebook/react-native/issues/34239#issuecomment-1234374559

Is this expected?

bouncer/src/repo/core/model/bson/repo_bson_sequence.cpp
```diff
@@ -21,12 +21,74 @@
...
+
+	builder.appendArray(REPO_SEQUENCE_LABEL_FRAMES, frames);
+
+	return builder.obj();
+}
+
+bool RepoSequence::isSizeOK()
+{
+	RepoBSON bson = *this;
+	if (bson.objsize() > REPO_MAX_OBJ_SIZE)
```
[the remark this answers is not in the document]
sebjf: Yes, done!

bouncer/src/repo/core/model/bson/repo_bson_ref.cpp
```diff
@@ -31,32 +33,149 @@ std::string RepoRef::convertTypeAsString(const RefType &type) {
...
+		}
+		auto typeStr = bson.getStringField(REPO_REF_LABEL_TYPE);
+		if (typeStr == REPO_REF_TYPE_FS)
+			type = RefType::FS;
 	}
-	else if (element.type() == ElementType::STRING) {
-		return element.String();
+	catch (repo::lib::RepoBSONException)
+	{
+		// A failure of the required properties should result in an empty node
```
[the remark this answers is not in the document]
sebjf: This is one of those cases where the calling logic explicitly handled the case where the ref node was empty, but it did print an error and return an empty stream... so not an expected case but one the previous code didn't want to be fatal. 

Looking into more, this block only gets called by `BlobFilesHandler`, and [here](https://github.com/3drepo/3drepobouncer/blob/5f559069381254d623f00875d424769e2b297c1f/bouncer/src/repo/core/handler/fileservice/repo_blob_files_handler.cpp#L73) it doesn't make any concessions for an empty response, so I would agree failing earlier is better here and have removed the try/catch, along with `IsEmpty()`.

bouncer/src/repo/core/model/bson/repo_node_model_revision.cpp
sebjf: 🙈 done!

bouncer/src/repo/core/model/bson/repo_node_metadata.cpp
sebjf: I thought the same but it does occur. Looking into it further the `AppendVisitor` filters out empty strings. I don't think this should be the behaviour, so I have removed that check, however existing uploads will still have the missing fields. 
If we add another type to the RepoVariant that can be null we may need to run a migration script in the future.

sebjf:
> It's struggling to compile the test for me: image
> 
> Quick google seems to think it's to do with requiring an env-var [facebook/react-native#34239 (comment)](https://github.com/facebook/react-native/issues/34239#issuecomment-1234374559)
> 
> Is this expected?

It is not expected, though I just tried to recompile and its failing for me too 🤨 - I have tweaked it so now it builds with a different method....

bouncer/src/repo/core/handler/repo_database_handler_mongo.cpp
```diff
@@ -208,25 +210,28 @@ repo::core::model::RepoBSON createRepoBSON(
...
-		else if (orgBson.hasLegacyFileReference()) {
-			std::unordered_map< std::string, std::pair<std::string, std::vector<uint8_t>> > binMap;
-			std::vector<std::pair<std::string, std::string>> extFileList = orgBson.getFileList();
-			for (const auto &pair : extFileList)
-			{
-				auto fileManager = fileservice::FileManager::getManager();
+		else if (orgBson.hasField(REPO_LABEL_OVERSIZED_FILES)) {
+			// The _extRef support has been deprecated, and collections should have been
+			// updated by the 5.4 migration scripts.
+			// In case we have any old documents remaining, this snippet can read the data,
```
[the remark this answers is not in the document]
sebjf: OK if its gone from .io then I'll remove it completely too!

I have run the 5.4 migration script on the tests database dump, though not pushed yet as this will break the tests for the other branches.

After doing this I realised that the 5.4 migration script didn't write the `size` parameter to the new `_blobRefs`, so I have created an additional migration script which infers it based on the elements, and used that to update the test's dump. We will need to run it against the database for bouncer to load legacy files for real though.

carmenfan:
> Hi @carmenfan,
> 
> The above comments have been addressed now, and the tests are passing.
> 
> Two things that have come up:
> 
> 1. **_blobRef.buffer.size**
> 
> I removed `_extRef` support and updated the tests. In doing so, I realised that the 5.4 migration script did not write the `_blobRef.buffer` `size` field. This field is required by bouncer because it reads the entire file in `BlobFilesHandler`. I wrote an additional migration script which sets this field, based on what's in `elements`. I have used the script to update the `tests` database, but not committed the script itself anywhere. If we want bouncer to read legacy nodes (e.g. to run `genStash`) then the script will need to run on production. If not though we don't need to do anything further.
> 
> Let me know what you'd like to do.
> 
> In either case, the revised tests database is on a dedicated [branch](https://github.com/3drepo/tests/issues/1), as it will break the tests on older versions of bouncer due to the missing `_extRef` nodes. We will need to merge this & remove the branch command from `travis.yml` before merging this PR.
> 
> 2. **Empty Strings**
> 
> The `RepoBSONBuilder::appendRepoVariant` method did not append empty strings, meaning some `metadata` entries will be missing keys. I have updated `RepoBSONBuilder` & the unit tests, and removed this condition from `MetadataNode::deserialise` so bouncer now always populates the key field, and assumes it is always populated.
> 
> Bouncer will not be able to read legacy MetadataNodes, if they have missing fields. Again this is only important if we want to run, e.g. `genStash`.
> 
> I have not written a migration script, but it should be simple to do. Let me know what you'd like to do.

@sebjf I think that's fine. I don't see us re-running gen stash on old data. 👍

carmenfan (dismissed):
Looks good. If you want to undo the change on travis.yml and merge in the dump in the test, feel free to merge everything!

carmenfan (approved):

PR #727, opened 2025-01-07
Issue 588 - Don't reset flags between stages

### LINKED ISSUE #588: model uploads should not be resetting the incomplete flag in between phases
### Description
currently, after each stage of processing, we reset the incomplete flag.

This causes a situation where if the process timeout/crash in a specific timing, the revision will look like it has been processed successfully.

We should stop resetting the flag in between to ensure this doesn't happen

### Steps to replicate
- upload a model
- force the process to exit in between process

### Current Behaviour
- it will return failure, but the user will be able to see a revision uploaded in the revision API

### Expected Behaviour
- the user should not be able to see the revision as a valid revision

sebjf:
This fixes #588

#### Description
This PR adjusts when the revision node state is updated, so that the COMPLETE flag is set only by calls that have the full context of the operations, rather than the individual stages. 

This means that flag updates are no longer necessarily symmetric or bracket any given operation (such as import, or tree generation, etc), but importantly a failure between stages will no longer be able to incorrectly flag an import as successful.

`updateRevisionStatus` has been added to `RepoController` for the purposes of `_generateStash`, since to update the status a database handler is needed, which comes from the worker pool managed by `RepoController`.


carmenfan (changes requested):
Looks fine, but unit tests failed 😆

carmenfan (approved):

PR #738, opened 2025-02-11
ISSUE #737 - Redefine elastic config validation

### LINKED ISSUE #737: bouncer_worker config shouldn't require elastic
- [x] bouncer_worker shouldn't require `elastic` to be defined in config to function
- [x] bouncer_worker also likely ignores `processMonitoring` in config

Charence:
This fixes #737

#### Description
- Validation wasn't allowing `elastic` to be undefined in the config as it recycled `envars` definition
- bouncer_worker can now be run without `elastic` being defined

#### Test cases
- Run bouncer_worker with `elastic` in config
- Run bouncer_worker without `elastic` in config

### DIFF
--- a/tools/bouncer_worker/src/lib/config.js
+++ b/tools/bouncer_worker/src/lib/config.js
@@ -57,12 +57,10 @@ const bouncer = object({
 });
 
 const elastic = object({
-	cloud: object({id: string()}).default(undefined),
-	auth: object({apiKey: string()}).default(undefined),
-	namespace: string(),
-	reload_connections: boolean(),
-	maxRetries: number(),
-	request_timeout: number(),
+	cloud: object({id: string().required()}),
+	auth: object({apiKey: string().required()}),
+	namespace: string().required(),
+	maxRetries: number().default(5),
 }).default(undefined);
 
 const logging = object({
--- a/tools/bouncer_worker/src/lib/config.js
+++ b/tools/bouncer_worker/src/lib/config.js
@@ -56,14 +56,22 @@ const bouncer = object({
 	log_dir: string(),
 });
 
+const elasticCloud = object({
+	id: string(),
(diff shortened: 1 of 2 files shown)


tools/bouncer_worker/src/lib/config.js
```diff
@@ -57,13 +57,13 @@ const bouncer = object({
 });
 
 const elastic = object({
-	cloud: envars,
-	auth: envars,
+	cloud: object({id: string()}).default(undefined),
```
carmenfan, line 60: I don't think you need the default(undefined) here, unless there's situation where someone would put in an elastic object and the application should still be functional without the cloud/auth definition.

In fact I think most of these fields should be required?

Things like maxRetries/reload_connections etc, unless the code checks for if they're empty and otherwise don't do anything, we want to impose a default?

carmenfan (changes requested):

tools/bouncer_worker/src/lib/config.js
carmenfan: Can we check through all the fields and determined if thye're required, optional, or should have default value?

Charence: I don't think the majority of these fields actually mean anything to Elasticsearch 7.17 :rofl: https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/7.17/basic-config.html

Charence: @carmenfan I've removed those that aren't supported. Do we care about adding

carmenfan: no, if we don't know about it then we probably don't pass it along i'm guessing?

carmenfan (changes requested):
linter is broken

carmenfan (approved):

PR #757, opened 2025-04-08
ISSUE #755 fix for appendElements using view that may be moved

### LINKED ISSUE #755: Hotfix: appendElementsUnique should make a new view when the document changes

sebjf:
This fixes #755

#### Description
Recreate the view of the builder that is being appended to if it changes - this is because a view is non-owning and so the builder may move the memory inside the loop.


sebjf:
@carmenfan done!

carmenfan (approved):

PR #761, opened 2025-04-11
Issue 760

### LINKED ISSUE #760: DWGs inside NWDs are missing text on import
Ensure TEXT entities inside NWDs, such as in this file: https://asitesol.sharepoint.com/:u:/s/3DRepoTeamFolder/EXll1U-ZyIdMkhlI8_AeEOMBdUha99Pr_NRW88eFZp5P-w?e=Yv9HWd are imported by ODA.

sebjf:
This fixes #760

**This PR is against 751 for the purposes of review. That PR should be approved first and merged, and then this one rebased to staging**

#### Description
This PR adds support for OdNwTextGeometry objects in the Navisworks importer. These objects are created when the Nwd contains a Dwg with TEXT entities (so far, this is the only case we've seen them encountered).

These objects exist in the NWD as parametric objects only. In order to get a tessellated version, we need to invoke the Gi functionality that ODA uses when rendering to a Gs device. This is achieved by instantiating an OdGiBaseVectorizer, which contains this functionality, and getting it to draw, much like we do for the other importers, but in this case we are only interested in the geometry.

In addition, a set of unit tests have been added for such objects.


test/src/unit/repo/manipulator/modelconvertor/import/odaImport/ut_repo_model_import_oda.cpp
```diff
@@ -59,6 +57,26 @@ namespace ODAModelImportUtils
 
 		return scene;
 	}
+
+	bool compareBounds(const repo::lib::RepoBounds& bounds, repo::lib::RepoVector3D min, repo::lib::RepoVector3D max, float tolerance)
+	{
+		return (bounds.min() - min).norm() < tolerance && (bounds.max() - max).norm();
```
FJThiel, line 63: Is it intentional that only the minimum is compared against the tolerance?

FJThiel (changes requested):
Neat. I only have trivial stuff to suggest.

FJThiel (approved):

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/file_processor_nwd.cpp
```diff
@@ -43,24 +43,41 @@ class OdExNwSystemServices : public RepoSystemServices
...
+	svcs = nullptr;
+}
+
 uint8_t FileProcessorNwd::readFile()
 {
 	int nRes = REPOERR_OK;
 
-	OdStaticRxObject<RepoNwServices> svcs;
-	odrxInitialize(&svcs);
+	bool shouldDeInitialise = createSharedSystemServices();
```
FJThiel, line 76: ~~I am not sure I understand this mechanism here. The pointer might not be null, because it is a static, so on a surface level it makes sense to me that individual calls to readFile only deinitialise what they initialised, but I see two problems here:~~
~~1. If readFile cannot be called in parallel, then this is always true and we don't need this mechanism.~~
~~2. If readFile can be called in parallel, then the call to readFile that initialised the pointer might finish before following calls and invalidate it while we are still using it.~~

~~Am I missing something?~~

Edit: Found the explanation in the header. Maybe add a one-line comment stating that this is to prevent an ODA bug on Linux and point to the full explanation in the header?

FJThiel (approved):
Looks good to me.

PR #765, opened 2025-05-14
Issue 764

### LINKED ISSUE #764: Change the AbstractModelImport superclass to use streaming API by default
Soon most of the importers will use the streaming import pattern, but the AbstractModelimport still has the old signature, even though this mostly throws an exception.

Once all the importer PRs are merged, clean this up.

FJThiel:
This fixes #764 

#### Description

With the new streaming approach to the importers we want to change the signature of AbstracModelImport and the classes that inherit from it. This mainly means changing modelImport from taking two parameters to three and the removal of generateRepoScene, since the scene object is becoming smaller and will be generated during the import process.

#### Test cases

All existing test cases are expected to pass.


sebjf:
@FJThiel  this is failing travis!

FJThiel:
@sebjf 
Travis is passing now that #769 has been merged in.

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_3drepo.h
```diff
@@ -73,26 +73,17 @@ namespace repo {
...
-				/**
-				* Generates a repo scene graph
-				* an internal representation needs to have
-				* been created before this call (e.g. by means of importModel())
-				* @return returns a populated RepoScene upon success.
-				*/
-				virtual repo::core::model::RepoScene* generateRepoScene(uint8_t &errMsg);
-
 				/**
 				* Import model from a given file.
```
sebjf, line 77: The description here is out of date!

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_abstract.h
```diff
@@ -47,28 +47,16 @@ namespace repo {
...
-				/**
-				* Generates a repo scene graph
-				* an internal representation needs to have
-				* been created before this call (e.g. by means of importModel())
-				* @return returns a populated RepoScene upon success.
-				*/
-				virtual repo::core::model::RepoScene* generateRepoScene(uint8_t &errMsg) = 0;
-
 				/**
 				* Import model from a given file
```
sebjf, line 51: Ditto here.

bouncer/src/repo/manipulator/modelconvertor/import/repo_model_import_manager.cpp
```diff
@@ -46,13 +46,12 @@ repo::core::model::RepoScene* ModelImportManager::ImportFromFile(
 		return nullptr;
 	}
 
-	repo::core::model::RepoScene* scene = nullptr;
 	repoTrace << "Importing model...";
-	if (modelConvertor->importModel(file, handler, error)) {
+	repo::core::model::RepoScene* scene = modelConvertor->importModel(file, handler, error);
+	if (scene) {
```
sebjf, line 51: This conditional will prevent line 69 from running, no? Also, it would be best not to set the error code to scene fail unless it has actually failed (line 53) - though that code is indeed not used unless scene is null (and we are also moving to handling errors via exceptions), that is an implementation detail so if it ever gets reported somewhere it would be quite confusing!

sebjf (approved):

Issue #30, opened 2015-10-08
[RepoClient] Utilise program_options for command line arguments


carmenfan:
At the moment command line arguments rely on being in the right order, and it is getting incredibly confusing with the complexity of different options available

It'll be a lot cleaner to use program_options and parse arguments that way. 

also dxrotate should probably be moved into config file.

PR #771, opened 2025-05-16
ISSUE #769 Import all geometry from Revit

### LINKED ISSUE #769: Create a default view that imports all Model elements for Revit
* Add an option to the import config to specify a view
* If the view is not set, create a default view that shows all the Model elements with no clipping etc.

sebjf:
This fixes #769 - import all geometry from Revit, not just that in the view

#### Description
This PR modifies the Revit importer to create a new 3D view, instead of trying to pick one from the file. The new View uses the Shaded (with Edges) view by default, and the options that control visibility are configured to filter as little as possible, meaning we should get all the 3D geometry in the file.

Users can still specify an existing view via the config file if desired. If this view cannot be found, or is not a suitable 3D view, then the importer will throw an exception with a suitable error code (it will *not* fall back to the default view).

This is set in the import config `view` member.

Users can also choose the display style for the default view. The non-default options are:

* `wireframe`
* `hiddenline`
* `shaded`

These are set in the import config json `style` member.

The styles correspond to the similarly named styles in Revit. The behaviour mimics the Revit UI options, so, choosing Shaded - which excludes lines - will also hide Surface Patterns.

Unit tests have been added for the new functionality.

Additionally, the following minor changes have been made:

1. The ModelImportConfig members have been made public and can now be initialised individually. The previous constructor is still present as an option, as it is used in a number of existing locations. This is done because the ModelImportConfig constructor was becoming unwieldly with the number of parameters. ModelImportConfig is part of the modelconvertor namespace (i.e. private), and has no internal logic, so there are more downsides to attempting encapsulation in this case than benefits.

2. The client initialisation has been updated to use ModelImportConfigs new public members directly rather than local variables, to make the code neater and reduce the chance of mis-assignment. This is not a replacement of https://github.com/3drepo/3drepobouncer/issues/30, which should still be considered in the future. 

4. The REPOERR_VALID_3D_VIEW_NOT_FOUND error has become REPOERR_VIEW_NOT_FOUND, and a new error REPOERR_VIEW_NOT_3D, is introduced for the case where a view is found but is not suitable. Given that users will be providing view names in a free entry field, it is likely important to disambiguate these cases.

5. Some additional methods have been added to SceneUtils to support the new unit tests.

**Textures and Tests**

This PR makes the texture lookups for Rvt all lowercase. When this PR is deployed, the Autodesk textures should all be renamed to be lowercase.

Windows (or at least NTFS) is case-insensitive, and since Autodesk applications are designed for Windows will also be, meaning we can't rely on correct capitalisation from ODA. The cleanest approach is to explicitly say we are working in lowercase on Linux.

If we add texture support to other Autodesk formats like NWDs, we must make sure they are also lowercase.

Additionally, this PR introduces two new test util functions (`setupTextures` and `unsetupTextures`) which set and clear the Rvt Textures environment variable, respectively. Since the return values of the system tests rely on whether this environment variable is set, the tests previously relied on a specific execution order, with one test in particular setting the variable. 

GTest does not enforce any particular execution order however, so this was only working coincidentally.

Each system test that uses the environment variable now sets or clears it at the appropriate time so they can be run in any order, or individually, and get the same result.

### DIFF
--- a/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/file_processor_nwd.cpp
+++ b/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/file_processor_nwd.cpp
@@ -73,6 +73,8 @@ uint8_t FileProcessorNwd::readFile()
 {
 	int nRes = REPOERR_OK;
 
+	// When running under 3drepobouncerClient, create a set of system services on demand
+	// for this one invocation. This is a workaround for a bug. See header for details.
 	bool shouldDeInitialise = createSharedSystemServices();
 	OdRxModule* pModule = ::odrxDynamicLinker()->loadModule(sNwDbModuleName, false);
 	try


sebjf:
This PR is flagged invalid until 762 is merged. We also need to decide before final merge if we want to import lines or not.

sebjf:
Once this is approved, https://github.com/3drepo/AssetGenerator/issues/17 should be merged in sync with this going into staging.

bouncer/src/repo/error_codes.h
carmenfan, line 1: We need to update IO to recognise the new error code (and revised error codes, if applicable)

Just noticed we didn't add it over there when we introduced 39 🙈 

Can you create an issue on io and update it as appropriate? https://github.com/3drepo/3drepo.io/blob/master/backend/src/v5/models/modelSettings.constants.js#L22

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/file_processor_rvt.cpp
```diff
@@ -303,6 +279,73 @@ void setupViewOverrides(OdBmDBViewPtr pView)
...
+	}
+	else if (style == "shaded") {
+		pDBView3d->setDisplayStyle(OdBm::ViewDisplayStyle::Shaded);
+	}
+	else {
+		pDBView3d->setDisplayStyle(OdBm::ViewDisplayStyle::ShadedWithEdges);
+	}
+
+	pDBView3d->setDetailLevel(OdBm::ViewDetailLevel::Fine);
+	pDBView3d->setViewDiscipline(OdBm::ViewDiscipline::Architectural); // Architectural shows all elements (search "About the View Discipline" in the Revit docs)
```
carmenfan, line 309: More of a sanity check/question than request for change: I guess we're still hiding annotations as per #767 because we filter them out further down the line?

carmenfan (commented):
Generally looks fine, just going to compile it and test it (also checking with Jozef whether we need to support the view immediately (if so i need to allocate some web dev time before we can merge this..)

There'so one thing to update on 3drepo.io if you don't mind doing it 🙈 and also a question for sanity check

PR #774, opened 2025-05-28
ISSUE #773 support for the world orientation in nwd partitions

### LINKED ISSUE #773: NWD importer should respect the World Orientation setting
The [World Orientation](https://help.autodesk.com/view/NAV/2025/ENU/?guid=GUID-698C9B45-4C49-4368-8EE6-F4FA47E79B2B) is a setting that can be made in Navisworks files. It is available via the OdNwPartition object. We should use this setting to ensure that NWDs with manually configured world orientations are correctly imported.

- [x] Create some test files to see if NWDs support multiple world orientations per partition and what they look like if so
- [x] Extract the world orientation(s) from the OdNwPartition
- [x] Apply the orientations to the geometry
- [ ] Make sure bounding boxes work
- [ ] Create unit tests

sebjf:
This fixes #773 - support for World Orientation of Navis partitions

#### Description
This PR adds a stage to the Navis importer that checks for the Up and North vectors of OdNwPartitions, and uses them to create a top-level transformation that puts geometry into the correct project coordinate conventions (X-East, Y-North, Z-Elevation) on import.

In doing this, the PR also adjusts slightly how the file is processed, first enumerating the Model entries, then their root elements (Nodes). This pattern (that a database has multiple models, each with its own branch, instead of just one branch with ModelItems being child nodes) is now [the one recommended by ODA](https://docs.opendesign.com/bimnv/bimnv_extract_nwd_nwc_data.html), and from 26.xx will be the only one supported.

Models (and the OdNwPartitions at their roots) correspond to the files/links in a Navis federation, so in theory can have their own World Transforms and Units and Transforms. In practice, we will only ever have one, because when saving NWDs, Navis bakes the federation under a new node that represents the NWD itself. Still, for completeness the PR is written such that multiple models could be supported.

Finally, unit tests are introduced for new functionality, and some unit tests have been accordingly updated to be more generic and verbose.

### DIFF
--- a/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_nwd.cpp
+++ b/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_nwd.cpp
@@ -164,6 +164,7 @@ struct RepoNwTraversalContext {
 	repo::manipulator::modelutility::RepoSceneBuilder* sceneBuilder = nullptr;
 	std::shared_ptr<repo::core::model::TransformationNode> parentNode;
 	std::unordered_map<std::string, repo::lib::RepoVariant> parentMetadata;
+	OdGeMatrix3d transform;
 	OdVectorizerPtr vectorizer;
 };
 
@@ -716,7 +717,7 @@ OdResult processGeometry(OdNwModelItemPtr pNode, RepoNwTraversalContext context)
 		if (!(*itFrag).isNull())
 		{
 			OdNwFragmentPtr pFrag = OdNwFragment::cast((*itFrag).safeOpenObject());
-			OdGeMatrix3d transformMatrix = pFrag->getTransformation();
+			OdGeMatrix3d transformMatrix = context.transform * pFrag->getTransformation();
 			OdNwObjectId geometryId = pFrag->getGeometryId(); // Returns an object ID of the base class OdNwGeometry object with geometry data.
 
 			if (geometryId.isNull() || !geometryId.isValid())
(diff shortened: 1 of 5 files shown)


bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_nwd.cpp
```diff
@@ -953,18 +954,79 @@ OdResult traverseSceneGraph(OdNwModelItemPtr pNode, RepoNwTraversalContext conte
...
+	for (auto& model : models)
+	{
+		OdNwModelPtr pModel = model.safeOpenObject();
+		OdNwModelItemPtr pModelItemRoot = pModel->getRootItem().safeOpenObject();
 		context.layer = pModelItemRoot;
-		context.parentNode = context.sceneBuilder->addNode(RepoBSONFactory::makeTransformationNode({}, "rootNode"));
-		context.vectorizer = OdVectorizer::createObject(pNwDb);
+
+		// Each linked file (Model) can have its own World Orientation and Units and
+		// Transformations set. In practice though, when saving an NWD/NWC (the only
```
FJThiel, line 1023: Just checking my understanding here:
In theory we could have multiple models with their own World Orientation etc., hence the loop, but in reality there is usually a root node introduced by Navis and the models baked under it, so there will only be one element in models. Is that correct like that?

FJThiel (commented):
Looks good to me. I would just like a sanity check (see comment) to make sure I understood correctly what happens.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_nwd.cpp
FJThiel: Got it!

FJThiel (approved):

FJThiel (approved):

PR #785, opened 2025-07-15
ISSUE #784 All metadata nodes should be children of their mesh siblings

### LINKED ISSUE #784: All metadata nodes should be children of their mesh siblings
Metadata nodes must always be accessible (i.e. children of) mesh nodes to which they are associated.

This is because the frontend uses them to lookup metadata, by mesh id.

- [x] Refactor scene utils and ensure there are no regressions in the test code
- [x] Make unit test pattern to check that all metadata nodes have their sibling meshnodes as parents also
- [x] Add this unit test to all file types that support metadata 
    - [x] IFC
    - [x] NWD
    - [x] RVT
    - [x] DWG
    - [x] DGN
    - [x] SPM
    - [x] BIM

sebjf:
This fixes #784

This PR adds a processing stage that ensures any metadata within leaf nodes is hooked up.

Leaf nodes are nodes that have only unnamed meshes as children. For these nodes, any metadata parented to the transform should also be parented to the meshes. This is assumed by the backend when it attempts to resolve external ids from metadata, based on the shared ids passed by the viewer (which will be those of possibly hidden mesh nodes, only).

The use of a processing pass is the best approach because the importers have varying degrees of context. Further if the tree logic can adopt the responsibility of identifying what is and is not a composite object in the future, the processing pass can be removed.

A unit test has been added that can that checks that metadata is correctly connected within all leaf nodes in a scene, and this has been added to the tests of all importers that support metadata (all of them except assimp).

The scene test utils have been updated to support this test, which includes making some APIs more explicit. As the previous graph tests did not include this inheritance, the findNodesByMetadata method is now implicitly "find *leaf* nodes by metadata", which reflects its original intent and avoids any changes to the existing tests.

### DIFF
--- a/bouncer/src/repo/core/handler/database/repo_query.h
+++ b/bouncer/src/repo/core/handler/database/repo_query.h
@@ -22,6 +22,7 @@
 #include "repo_query_fwd.h"
 
 #include <vector>
+#include <set>
 #include <string>
 #include <memory>
 #include <variant>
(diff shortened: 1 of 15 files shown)


carmenfan (approved):
Tested IFC and RVT and they seem to be working.

PR #794, opened 2025-07-29
Issue 793 - Update googletest

### LINKED ISSUE #793: Update googletest
The current version of googletest is built with/for cmake 2.8. CMake 4 rc deprecates 3.5 features so will not build this project anymore without overrides.

We should upgrade googletest to ensure bouncer can be built with cmake 4 before cmake 3 is deprecated entirely.

sebjf:
This fixes #793

#### Description

This PR simply updates the version of Googletest to the latest, as the current version is quite old and we are going to get compatibility errors with cmake soon.


sebjf:
@FJThiel, this is a dependency update on the tests so you can just look at the code - no need to test it yourself if you don't want to!

PR #799, opened 2025-08-06
Issue 775 - Fix Unicode Support

### LINKED ISSUE #775: All importers should support UTF-8 filenames
Import instructions are delivered to bouncer via the import file config. This being Json is by definition UTF8 encoded. All importers should support Unicode at least until the point of opening the file (identifying support for Unicode metadata and tree entries is out of scope).

For context, Windows has its own codepage (ANSI) which is the system default. Supporting UTF8 basically means going through every string conversion method (e.g. from std::string to Odstring, or from char* to boost::filesystem::path, etc) and making sure if the byte sequence is reinterpreted, it is reinterpreted as UTF8...

sebjf:
This fixes #775

#### Description
This PR makes a number of minor changes in order to fully support Unicode filenames. Specifically, boost filesystem has been replaced with the C++ standard filesystem, and the paths to explicitly specify utf8 encoding (since we get filenames from Json).

Additionally, the locale of the executables (client & test) is set to the system default on start-up. This is necessary for ODA to work with UTF-8 strings.


sebjf:
@carmenfan, requesting your review when you're back as Felix is on holiday 😄

PR #803, opened 2025-08-27
Issue 802 Replace svg converter

### LINKED ISSUE #802: Replace svg converter
### Description
Product issue: [#872](https://github.com/3drepo/3D-Repo-Product-Team/issues/872)

This task is to change the bouncer worker to call a different tool for converting pdf to svg. Previously, Inkscape was used, but it was found to have horrid performance with big pdfs (see product ticket).

### Goals

- [ ] As a user, I want to put complex pdfs through bouncer without it going out of memory.

### Tasks

- [ ] Change the bouncer worker script to call a different tool for the conversion.

### Related Resources

MockUp : `link to mockup`

FJThiel:
This fixes #802 

#### Description

This task is to change the bouncer worker to call a different tool for converting pdf to svg. Previously, Inkscape was used, but it was found to have horrid performance with big pdfs (see product ticket). This PR now changes it to use [pdf2svg](https://github.com/dawbarton/pdf2svg).

### DIFF
--- a/tools/bouncer_worker/src/tasks/imageProcessing.js
+++ b/tools/bouncer_worker/src/tasks/imageProcessing.js
@@ -29,7 +29,10 @@ ImageProcessing.testImageClient = async () => {
 	try {
 		await run(PDF2SVG, [], { label: 'INIT' });
 	} catch (err) {
-		throw new Error(`Failed to call ${PDF2SVG}: ${err?.message}`);
+		// Check for error codes indicating that the tool is not installed
+		if(err == 127 || err == 9009){
+			throw new Error(`Failed to call ${PDF2SVG}: ${err?.message}`);
+		}
 	}
 };
 
--- a/tools/bouncer_worker/src/tasks/imageProcessing.js
+++ b/tools/bouncer_worker/src/tasks/imageProcessing.js
@@ -36,7 +36,7 @@ ImageProcessing.testImageClient = async () => {
 ImageProcessing.generateSVG = async (file, output, taskInfo) => {
 	const retVal = await run(
 		PDF2SVG,
-		[file,output,'all'],
+		[file, output, 'all'],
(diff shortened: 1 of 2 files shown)


tools/bouncer_worker/src/tasks/imageProcessing.js
```diff
@@ -23,28 +23,28 @@ const {
 } = require('../constants/errorCodes');
 
 const ImageProcessing = {};
-const INKSCAPE = 'inkscape';
+const PDF2SVG = 'pdf2svg';
 
 ImageProcessing.testImageClient = async () => {
 	try {
-		await run(INKSCAPE, ['--version'], { label: 'INIT' });
+		await run(PDF2SVG, [], { label: 'INIT' });
```
carmenfan, line 30: @FJThiel This doesn't work on linux 
image

Seems like the exit code isn't 0 when you run it with no params on linux, so maybe we need to check the exit state isn't 127 or we need some other strat here.

carmenfan (changes requested):

tools/bouncer_worker/src/tasks/imageProcessing.js
FJThiel: Oh yes, I am sure it runs somewhere in the docker image, but I would prefer not to stick my fingers in there to find out whether I made a typo if I can help it 😅

Thanks for the readme, I think that will help me.

FJThiel: After checking the source of pdf2svg for the errors it can return (-1 to -5),   and reading a bit on what operating systems to with negative exit codes, I think the most robust way might be to include a tiny pdf file and have it convert that for testing. That way we get a 0 when all is working and can flag all errors as such.

@carmenfan, would that be agreeable?

carmenfan (dismissed):

PR #813, opened 2025-11-17
#811 commit to create PR

### LINKED ISSUE #811: Test bouncer workflow
Branch to run workflow test now ive removed pr branch logic

nkuberaf:
This fixes #811


.github/workflows/build-bouncer.yml
carmenfan: (I can't comment on the lines since they aren't changed:

I think you may as well remove these 2 as well:
`DOCKER_BUILDKIT` - Isn't it better to just do it on the docker line? i.e `DOCKER_BUILD_KIT=1 docker....`
`LICENSE_CHECK` - It's not even used, we're setting it to 1 on the docker line.

nkuberaf: removed license check and put env variable buildkit in docker part as it cant be parsed in the dockerfile but where it runs instead (the workflow runner)

changefile.txt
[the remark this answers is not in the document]
nkuberaf: done

PR #831, opened 2026-03-17
Issue 797 - Clash Detection Engine

### LINKED ISSUE #797: Clash Detection Engine
https://github.com/3drepo/3D-Repo-Product-Team/issues/639

sebjf:
This fixes #797 

This PR exists to build images at the moment. When it is ready the invalid label will be removed.


ifcUtils/repo_ifc_serialiser.cpp
```diff
@@ -828,15 +829,21 @@ bool IfcSerialiser::shouldGroupByType(const IfcSchema::IfcObjectDefinition* obje
...
 	{
 		name = object->declaration().name();
 	}
 
-	auto transform = repo::core::model::RepoBSONFactory::makeTransformationNode({}, name, { parentId });
+	// Create transformation node with the new transform
+	auto transform = repo::core::model::RepoBSONFactory::makeTransformationNode(transformMat, name, { parentId });
+
+	// Insert corrective transform for lookup
+	auto parentCorr = corrTransforms[parentId];
```
sebjf: @FJThiel, if I understand correctly, corrTransforms is effectively storing the world-to-local transform for each parent? Is this right, or is the combination of inverses actually storing something else?

If the former, is there a reason to store the world-to-local, instead of local-to-world & then applying the appropriate inverse when getting the difference between the ifc's intended transform and the current transformation nodes transform on line 1073?

I ask because its a little hard to follow! Perhaps just some renaming/additional comments would help.

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_hard.cpp
```diff
@@ -0,0 +1,331 @@
...
+	std::mutex clashesMutex{};
+	std::mutex cacheMutex{};
+
+	// Define thread behaviour
+	auto narrowPhaseThread = [&](std::queue<Narrowphase>& queue)
+		{
+			while (true) {
+
+				// If we are out of samples, we unlock and terminate
+				if (queue.empty()) {
```
sebjf: Can !queue.empty() not just be the condition of the while loop?

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_hard.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+				bool bInitialised = false;
+				{
+					// Shared lock on the object so we can check the status
+					std::shared_lock lock{ b->mutex };
+					bInitialised = b->isInitialised();
+				}
+
+				// If one or both them are not initialised, we are initialising them
+				// Needs a lock on both of them to prevent a deadlock situation with
```
sebjf: This is interesting... can you explain how this deadlock happens?

sebjf (commented):
Question about the potential deadlock in hard mode

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_hard.cpp
FJThiel: Done. [3323fc6](https://github.com/3drepo/3drepobouncer/pull/831/commits/3323fc60c7e4627ec2c2588f9a1a1751ea5447d3)

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_hard.cpp
FJThiel: Yeah... I see it too now.
Fixed in [6b5ac08](https://github.com/3drepo/3drepobouncer/pull/831/commits/6b5ac084203ea3f04d9ace2f7879fc9e411233a8)

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_clearance.cpp
```diff
@@ -0,0 +1,378 @@
...
+				{
+					// Shared lock on the object so we can check the status
+					std::shared_lock lock{ b->mutex };
+					bInitialised = b->isInitialised();
+				}
+
+				// If one or both them are not initialised, we are initialising them
+				// Needs a lock on both of them to prevent a deadlock situation with
+				// an edge in the opposite direction in another thread
+				if (!aInitialised || !bInitialised)
```
sebjf: Same thing there! Is the condition different for clearance?

sebjf (commented):
Similar comment about the deadlocks in clearance.

Overall, the multithreading implementation is very good! Most importantly it is easy to follow, and self-contained.

The only thing I wonder is about the caching. unique_ptr is not thread safe as you've spotted. Originally, this approach was taken because it allowed `getReference` to be called, and then the entry would _implicitly_ become reference counted.

If we must explicitly release the cache entries, I wonder if it is better to make the cache use shared_ptrs.

The implementation would become simpler: basically the shared_ptr would be to a record, and would own the record. The map would be between the keys and the shared_ptrs, and getReference would copy the shared_ptrs.

After the list of scheduled entries has been turned into shared_ptrs by getReference, a `finalise` method would have to be called on the cache to tell it clear the map and so drop its own references so they can be cleaned up.

This may or may not be easier to follow - I will leave it up to you to think about!

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_clearance.cpp
FJThiel: Fixed. [d61a0f8](https://github.com/3drepo/3drepobouncer/pull/831/commits/d61a0f8f540e00453289b9c5774013b36c597199)

sebjf:
@FJThiel just took a second look at clearance, and actually given that this requires the cache be freed in two places I think it is important the cache behaves as expected (i.e. it uses threadsafe smart pointers that dont have to be cleaned up explicitly).

bouncer/src/repo/core/handler/repo_database_handler_mongo.cpp
```diff
@@ -345,7 +340,8 @@ void MongoDatabaseHandler::createIndex(const std::string& database, const std::s
 			auto col = db.collection(collection);
 			auto obj = (repo::core::model::RepoBSON)index;
 
-			repoInfo << "Creating index for :" << database << "." << collection << " : index: " << obj.toString() << " sparse: " << sparse;
+			if(!suppressInfo)
```
carmenfan, line 343: Is there a reason why we need this flag?

Maybe downgrade it to repoVerbose/repoDebug if we only want this to appear for debugging?

bouncer/src/repo/core/model/bson/repo_node_mesh.h
```diff
@@ -323,6 +323,8 @@ namespace repo {
 				void removeDuplicateVertices();
 
 				static void transformBoundingBox(repo::lib::RepoBounds& bounds, repo::lib::RepoMatrix matrix);
+
+				static void transformNormals(std::vector<repo::lib::RepoVector3D>& normals, repo::lib::RepoMatrix matrix);
```
carmenfan: should the matrix be `const RepoMatrix &matrix`?

gives the intent that it's read only (esp when you're intending to overwrite the normal) and avoid the copy 

(but my C++ is very rusty so not sure about best practice)

bouncer/src/repo/core/model/bson/repo_node_revision.h
```diff
@@ -22,6 +22,7 @@
 #pragma once
 #include "repo_node.h"
 #include "repo/core/model/repo_model_global.h"
+#include <ctime>
```
carmenfan, line 25: Do we need this? there's no other change within this file (and it doesn't look like `repo_node_revision.cpp` was changed)

bouncer/src/repo/lib/datastructure/repo_range.cpp
```diff
@@ -0,0 +1,18 @@
+/**
```
carmenfan: do we need this file if it's empty?

test/src/unit/repo/lib/ut_repo_uuid.cpp
```diff
@@ -284,5 +287,53 @@ TEST(RepoUUIDTest, gteOperatorTest)
...
+	
+	bool collision = false;
+
+	auto threadBehaviour = [&]
+		{
+			for (int i = 0; i < numSamplesPerThread; i++)
+			{
+				// All threads abort once the first collision is found
+				if (collision)
+					return;
```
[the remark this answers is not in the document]
FJThiel: Changed in [eaca173](https://github.com/3drepo/3drepobouncer/pull/831/commits/eaca173b6d0e542f2299c583c63e3deecb27685a)

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_node_cache.cpp
FJThiel, line 1: This file appears to have no meaningful content. Maybe remove it?

bouncer/src/repo/core/handler/database/repo_query.h
```diff
@@ -62,7 +62,25 @@ namespace repo {
...
-						Eq(const char* field, std::vector<T> oneOf) :
+						Eq(std::string field, const std::set<T>& oneOf) :
+							field(field)
+						{
+							for (auto& v : oneOf) {
+								values.push_back(v);
+							}
+						}
+
+						template<typename T>
```
FJThiel, line 73: There is some minor duplication here. Could the methods using the char* call the std::string methods or the other way around?

bouncer/src/repo/lib/datastructure/repo_line.h
```diff
@@ -0,0 +1,69 @@
+/**
+*  Copyright (C) 2016 3D Repo Ltd
```
FJThiel: That should probably be 2025 or 2026, right?

bouncer/src/repo/lib/datastructure/repo_matrix.cpp
```diff
@@ -0,0 +1,488 @@
...
+	}
+}
+
+template<typename T>
+_RepoMatrix<T>::_RepoMatrix(const std::vector<float>& mat)
+	:_RepoMatrix()
+{
+	for (int i = 0; i < mat.size(); ++i)
+	{
+		if (i >= 16) break;
```
FJThiel, line 51: Might be nice to have a warning here. I don't like that this can just fail silently with a mangled/incomplete matrix.

Same for the other constructor below that is overloaded with double

bouncer/src/repo/lib/datastructure/repo_matrix.cpp
```diff
@@ -0,0 +1,488 @@
...
+	memset(data, 0, 16 * sizeof(T));
+	data[0] = (T)1;
+	data[5] = (T)1;
+	data[10] = (T)1;
+	data[15] = (T)1;
+}
+
+template<typename T>
+template<typename T2>
+repo::lib::_RepoMatrix<T>::_RepoMatrix(const T2* coefficients, bool rowMajor)
```
FJThiel, line 35: This is the only constructor with a flag for row major / column major. Is it meant to say that the resulting matrix is to be in that format or that the incoming coefficient data is in that format?

FJThiel: After reading the rest, I have the impressions that we consider all our matrices to be row major. Is that correct?

bouncer/src/repo/lib/datastructure/repo_range.cpp
FJThiel, line 1: This cpp file is empty. Do we need it?

bouncer/src/repo/lib/datastructure/repo_triangle.cpp
```diff
@@ -0,0 +1,33 @@
+/**
+*  Copyright (C) 2016 3D Repo Ltd
```
FJThiel: Old year

bouncer/src/repo/lib/datastructure/repo_triangle.h
```diff
@@ -0,0 +1,79 @@
...
+				auto ab = b - a;
+				auto ac = c - a;
+				auto n = ab.crossProduct(ac);
+				n.normalize();
+				return n;
+			}
+
+			const repo::lib::_RepoVector3D<T>& operator[](size_t i) const
+			{
+				return reinterpret_cast<const repo::lib::_RepoVector3D<T>*>(this)[i];
```
FJThiel: If I understand this correctly, this will only work as expected for 0 <= i < 3?
Maybe we should add a guard here for anything out of that range.

bouncer/src/repo/lib/repo_units.h
```diff
@@ -0,0 +1,55 @@
+/**
+*  Copyright (C) 2015 3D Repo Ltd
```
FJThiel: Old year in header

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_dgn.cpp
```diff
@@ -106,9 +106,8 @@ bool DataProcessorDgn::doDraw(OdUInt32 i, const OdGiDrawable* pDrawable)
 	collector->pushDrawContext(ctx.get());
 	auto ret = OdGsBaseMaterialView::doDraw(i, pDrawable);
 	collector->popDrawContext(ctx.get());
-
-	auto meshes = ctx->extractMeshes();
-	if (meshes.size())
+	
+	if (ctx->hasMeshes())
```
FJThiel, line 110: In data_processor_dgn.cpp, we check for ctx first before we check whether it has meshes. Maybe we could make this consistent with each other.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.cpp
```diff
@@ -160,12 +173,31 @@ void GeometryCollector::Context::setMaterial(const repo_material_t& material)
...
+}
+
+std::vector<std::pair<repo::core::model::MeshNode, repo_material_t>> GeometryCollector::Context::extractMeshes(const repo::lib::RepoMatrix& m)
 {
 	std::vector<std::pair<repo::core::model::MeshNode, repo_material_t>> pairs;
 	for (auto& p : meshBuilders) {
 		std::vector<repo::core::model::MeshNode> meshes;
-		p.second->extractMeshes(meshes);
+		p.second->extractMeshes(meshes, m);
 		for (auto& m : meshes)
```
FJThiel: Double use of the variable name m: once for the parameter and once for the meshes in the loop.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/geometry_collector.h
```diff
@@ -174,7 +180,16 @@ namespace repo {
 							));
 						}
 
-						std::vector<std::pair<repo::core::model::MeshNode, repo::lib::repo_material_t>> extractMeshes();
+						/*
+						* Turns the mesh_data entries into a set of MeshNodes, each paried with a
```
FJThiel:
```suggestion
						* Turns the mesh_data entries into a set of MeshNodes, each paired with a
```

bouncer/src/repo/manipulator/modelutility/clashdetection/bvh_operators.cpp
```diff
@@ -0,0 +1,351 @@
...
+			pairs.push({ left.first_child_or_primitive + 0, right.first_child_or_primitive + 1 });
+			pairs.push({ left.first_child_or_primitive + 1, right.first_child_or_primitive + 0 });
+			pairs.push({ left.first_child_or_primitive + 1, right.first_child_or_primitive + 1 });
+		}
+	}
+}
+
+void Traversal::operator()(const bvh::Bvh<double>& a) {
+	// The goal of an internal traversal is to find all pairs in a single tree
+	// that overlap eachother. The traversal operates on pairs of siblings.
```
FJThiel:
```suggestion
	// that overlap each other. The traversal operates on pairs of siblings.
```

bouncer/src/repo/manipulator/modelutility/clashdetection/bvh_operators.cpp
```diff
@@ -0,0 +1,351 @@
...
+		};
+		result.max = {
+			std::min(a.max[0], b.max[0]),
+			std::min(a.max[1], b.max[1]),
+			std::min(a.max[2], b.max[2])
+		};
+		return result;
+	}
+
+	double distance2(const bvh::BoundingBox<double>& a, const bvh::BoundingBox<double>& b)
```
FJThiel, line 202: Does the 2 have a particular meaning or was there a "distance(...)" in the file previously that has now been removed?

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_pipelines_utils.cpp
```diff
@@ -0,0 +1,50 @@
...
+*  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+*  GNU Affero General Public License for more details.
+*
+*  You should have received a copy of the GNU Affero General Public License
+*  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "clash_pipelines_utils.h"
+
+#include "hopscotch-map/bhopscotch_map.h"
```
FJThiel: I don't think this is used in this file?

bouncer/src/repo/manipulator/modeloptimizer/bvh/single_ray_traverser.hpp
```diff
@@ -24,6 +24,7 @@ SOFTWARE.
 #define BVH_SINGLE_RAY_TRAVERSAL_HPP
 
 #include <cassert>
+#include <optional>
```
carmenfan, line 27: Any reason why this needs to be included here ? no change of code for the rest of the file

bouncer/src/repo/lib/hopscotch_map/bhopscotch_map.h
carmenfan, line 1: Tiny thing, but since kebab case isn't used in our file naming conventions, could we rename `hopscotch-map`?

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_clearance.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+	result.positions = {
+		static_cast<const ClearanceClash&>(clash).line.start,
+		static_cast<const ClearanceClash&>(clash).line.end
+	};
+
+	size_t hash = 0;
+	std::hash<double> hasher;
+	for (auto& p : result.positions) {
+		hash ^= hasher(p.x) + 0x9e3779b9 + (hash << 6) + (hash >> 2);
```
carmenfan: Can we refactor this a bit to reduce duplication (the 3 lines are identical) - perhaps right a helper function that just takes a number?


It'd also be helpful to put this in a constant `0x9e3779b9` so it helps self documenting - I had no idea why this number until I asked chatGPT, so perhaps if we can call it `HASH_GOLDEN_RATIO` or something it'll better explain the intention?

Or maybe just a comment on saying this is effectively `hash_combine` from boost library would also suffice.

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_exceptions.h
```diff
@@ -0,0 +1,114 @@
...
+#include <set>
+
+#include <repo/lib/datastructure/repo_container.h>
+
+namespace repo {
+	namespace manipulator {
+		namespace modelutility {
+			namespace clash {
+
+				struct ClashDetectionException
```
carmenfan: It's a bit odd to define exceptions that doesn't inherit from stdException (or RepoException in lib)

I feel like whenever a class is named as an Exception it is expected to inherit from the base exception class and carries the behaviour they do. Is there a reason why we don't inherit it? Perhaps we should rename them to `Errors` instead to avoid confusion?

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_hard.cpp
carmenfan, line 1: I think in general there's quite a bit of duplication between this and clash_clearance? which makes sense since they're just different conditions.

Would it be possible to move some of these common behaviour into the parent class?

For e.g. `createClashReport` could be refactored to be at the parent level, with an inherited function to pass the list of positions. I feel like the `run` function also could be tweaked similarly.

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_node_cache.cpp
```diff
@@ -0,0 +1,20 @@
...
+*  This program is distributed in the hope that it will be useful,
+*  but WITHOUT ANY WARRANTY; without even the implied warranty of
+*  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+*  GNU Affero General Public License for more details.
+*
+*  You should have received a copy of the GNU Affero General Public License
+*  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "clash_node_cache.h"
```
carmenfan: do we need a cpp file if theres' no defintiions?

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_pipelines.h
```diff
@@ -0,0 +1,128 @@
...
+					// From mesh node unique id to an index into the meshes array - same as the
+					// one the bvh uses.
+
+					std::unordered_map<repo::lib::RepoUUID, size_t, repo::lib::RepoUUIDHasher> uuidToIndexMap;
+
+					Graph(std::vector<Node> nodes);
+
+					Node& getNode(size_t primitive) const;
+
+					Node& getNode(const repo::lib::RepoUUID& uniqueId) const;
```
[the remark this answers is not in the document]
sebjf: getNode casts away const as the intended design is that the vector does not change, however it is necessary for Node instances to self-modify (such as when being initialised).

test/src/unit/repo/core/model/bson/ut_repo_bson_factory.cpp
```diff
@@ -325,7 +320,7 @@ TEST(RepoBSONFactoryTest, MakeTransformationNodeTest)
 	EXPECT_EQ(name, trans2.getName());
 	auto matrix = trans2.getTransMatrix();
 
-	EXPECT_TRUE(compareStdVectors(transMatFlat, matrix.getData()));
+	EXPECT_THAT(memcmp(transMatFlat.data(), matrix.getData(), 16), Eq(0));
```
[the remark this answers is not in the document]
sebjf: Size argument has been updated. In this case the matrices must be bit-identical so the use of memcmp is correct.

test/src/unit/repo/manipulator/modelutility/clashdetection/ut_repo_bvh_operators.cpp
```diff
@@ -0,0 +1,151 @@
...
+	clashGenerator.downcastVertices = false;
+
+	auto hyperplanes = std::vector<repo::lib::RepoVector3D64>{
+		repo::lib::RepoVector3D64(1, 0, 0),
+		repo::lib::RepoVector3D64(0, 1, 0),
+		repo::lib::RepoVector3D64(0, 0, 1)
+	};
+
+	for (int itr = 0; itr < 100000; ++itr) {
+
```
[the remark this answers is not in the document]
sebjf: Agreed but out of scope for the moment since we don't have soak test infrastructure

test/src/unit/repo/manipulator/modelutility/clashdetection/ut_repo_bvh_operators.cpp
```diff
@@ -0,0 +1,151 @@
...
+				EXPECT_THAT(bvh::predicates::contacts(createBounds(a), createBounds(b)), IsFalse());
+			}
+
+			a = a + (d.end - d.start);
+
+			EXPECT_THAT(bvh::predicates::contacts(createBounds(a), createBounds(b)), IsTrue());
+		}
+	}
+
+	for (int itr = 0; itr < 100000; ++itr) {
```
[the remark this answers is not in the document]
sebjf: Agreed but out of scope for the moment since we don't have soak test infrastructure

bouncer/src/repo/lib/datastructure/repo_bounds.cpp
```diff
@@ -67,4 +93,88 @@ const RepoVector3D64& RepoBounds::max() const
...
+}
+
+bool RepoBounds::operator>(const RepoBounds& other) const
+{
+	return size().norm() > other.size().norm();
+}
+
+repo::lib::RepoBounds RepoBounds::empty()
+{
+	return RepoBounds(repo::lib::RepoVector3D64(0, 0, 0), repo::lib::RepoVector3D64(0, 0, 0));
```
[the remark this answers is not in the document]
sebjf: Fortunately this member is unused so can be removed.

bouncer/src/repo/lib/datastructure/repo_bounds.cpp
```diff
@@ -67,4 +93,88 @@ const RepoVector3D64& RepoBounds::max() const
...
+	max.y = std::nexttoward(max.y, std::numeric_limits<double>::infinity());
+	max.z = std::nexttoward(max.z, std::numeric_limits<double>::infinity());
+
+	return repo::lib::RepoBounds(min,max);
+}
+
+REPO_API_EXPORT repo::lib::RepoBounds repo::lib::operator*(const _RepoVector3D<double>& scalars, const repo::lib::RepoBounds& bounds)
+{
+	return { bounds.min() * scalars, bounds.max() * scalars };
+}
```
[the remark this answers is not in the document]
sebjf: The repovector constructor (which the curly brace initialisation should call in this case) reorders min/max pointwise as necessary. A regression test has been added to check this.

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_node_cache.h
```diff
@@ -0,0 +1,134 @@
...
+					// the associated entry is removed, even if other entries are emplaced or
+					// erased.
+
+					std::unordered_map<const Key*, std::unique_ptr<Record>> map;
+
+					~ResourceCache() noexcept(false) {
+						if(map.size()) {
+							throw std::runtime_error("ResourceCache destroyed while still holding records");
+						}
+					}
```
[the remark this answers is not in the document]
sebjf: The new error handling logic replies on nested exceptions for reporting. In this case we would terminate the process anyway, so leave this in place.

bouncer/src/repo/manipulator/modelutility/clashdetection/geometry_exceptions.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace geometry
+{
+	/*
+	* An exception thrown to indicate an unrecoverable logic error during a
+	* geometry test. Such exceptions are thrown when the process state itself is
+	* fine, but the test cannot proceed & provide sensible return values. Whatever
+	* algorithm is relying on the result should abort.
+	*/
+	struct GeometryTestException : public std::runtime_error
```
carmenfan, line 31: Is there a reason why this is its own class vs the clash_exceptions?

bouncer/src/repo/manipulator/modelutility/spatialpartitioning/repo_spatial_partitioner_rdtree.cpp
```diff
@@ -18,6 +18,7 @@
 #include "repo_spatial_partitioner_rdtree.h"
 #include "../../../core/model/bson/repo_node_mesh.h"
 #include "../../../core/model/bson/repo_node_supermesh.h"
+#include <repo_log.h>
```
carmenfan, line 21: really wondering if i'm missing something 🙈 

Is it like a dependency is using this?

bouncer/src/repo/manipulator/modelutility/repo_clash_detection_config.cpp
```diff
@@ -0,0 +1,217 @@
...
+	CompositeObjectMap mapB;
+
+	ClashConfigParser(ClashDetectionConfig& config)
+		:config(config)
+	{
+		parsers["type"] = new ClashTypeParser(config.type);
+		parsers["tolerance"] = new NumberParser<double>(config.tolerance);
+		parsers["numThreads"] = new NumberParser<int>(config.numThreads);
+		parsers["resultsFile"] = new StringParser(config.resultsFile);
+		parsers["setA"] = new ArrayParser(new CompositeObjectSetParser(this, mapA));
```
carmenfan, line 165: I'm not sure whether we want to do anything here, but just wanted to point out an issue @ChristosTsiotsias had when he was implementing the web side of the feature.

He spent a few days wondering why his clash runs were returning no results when the were clearly some within the models... it turns out he got the schema wrong and thought `setA` and `setB` takes in an object instead of an array of objects.

Instead of erroring, the code here just parsed it into empty arrays, thus nothing to clash.

This got me thinking:
1. Should we throw an error from the bouncer side when the config schema is incorrect?
2. Should we throw an error from the bouncer side if the clash run was given no compsite ID entries?

bouncer/src/repo/manipulator/modelutility/repo_json_parser.h
```diff
@@ -0,0 +1,424 @@
...
+*  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+*  GNU Affero General Public License for more details.
+*
+*  You should have received a copy of the GNU Affero General Public License
+*  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#pragma once
+
+#include "repo/manipulator/modelutility/rapidjson/reader.h"
```
carmenfan: I feel like this file and both the rapidjson folder are living in the wrong folder? it's less related to model utility but a general library to read json into repo data structures? (which live the `lib` folder)

so I feel like this and rapidjson should also live over there - they're generic utilities.

but I can see that rapidjson weren't introduced in this branch, so feel free to ignore me 😆

bouncer/src/repo/manipulator/modelutility/repo_mesh_map_reorganiser.cpp
```diff
@@ -16,6 +16,7 @@
 */
 #include "repo_mesh_map_reorganiser.h"
 #include "repo/core/model/bson/repo_bson_factory.h"
+#include <repo_log.h>
```
carmenfan, line 19: ☝️ 

If these are genuinely needed, just ignore me!

tools/bouncer_worker/src/lib/config.js
```diff
@@ -115,6 +116,8 @@ const init = () => {
 			config.taskLogDir = config.rabbitmq.sharedDir;
 		}
 
+		Config.replaceSharedDirTag = (data) => data.replace('$SHARED_SPACE', config.rabbitmq.sharedDir);
```
carmenfan: it's a bit of a weird place to declare a function declaration that isn't within the function scope or require the states within the function.

Unless i'm missing something, I think we can just move it out to here? 

Image

tools/bouncer_worker/src/lib/processParams.js
```diff
@@ -27,8 +27,8 @@ const parseParameters = () => {
 		describe: 'exit upon finishing the defined amount of tasks. Queue must also specified.',
 		number: true,
 	}).option('queue', {
-		describe: 'specify which queue to run on [job|model|drawing]',
-		choice: ['job', 'model', 'drawing'],
+		describe: 'specify which queue to run on [job|model|drawing|clash]',
+		choice: ['job', 'model', 'drawing', 'clash'],
```
carmenfan: not your fault as you didn't write this (probably mine 😆 )

Unless we introduce cyclic dependency (in which case we should probably refactor it into a constant file), it would be better if we just take it at run time from the declaration in queueHandler.js (will have to export it):

```suggestion
		choice: Object.values(queueLabel),
```

carmenfan: Similar with the line above

tools/bouncer_worker/src/queues/clashQueueHandler.js
```diff
@@ -0,0 +1,78 @@
...
+	} catch (err) {
+		logger.error(`Error running clash detection: ${err.message || err}`, logLabel);
+		returnMessage.value = err;
+		returnMessage.message = err.message || err;
+	}
+
+	callback(JSON.stringify(returnMessage));
+};
+
+Handler.validateConfiguration = (label) => callbackQueueSpecified(label)
```
carmenfan: curious as to why you need this?

We have a yup validator in the config.js, I'd assume that would've flagged if any of these are missing?

Similar to what i'm saying to christos on https://github.com/3drepo/3drepo.io/pull/5960, it might also be worth expanding the validator to do check the the file paths exists and the application have sufficient access rights.

tools/bouncer_worker/src/queues/drawingQueueHandler.js
```diff
@@ -68,9 +69,10 @@ Handler.onMessageReceived = async (cmd, rid, callback) => {
 
 	const returnMessage = {
 		value: ERRCODE_OK,
-		database,
-		project: model,
+		teamspace: database,
+		container: model,
```
carmenfan, line 69: @ChristosTsiotsias  are you aware of this change? I think it might break current io behaviour

tools/bouncer_worker/src/queues/clashQueueHandler.js
```diff
@@ -0,0 +1,78 @@
...
+
+	const returnMessage = {
+		value: ERRCODE_OK,
+		results: resultsFile,
+		teamspace,
+		project,
+		type: MSGTYPE_CLASH,
+	};
+
+	try {
```
carmenfan, line 59: Can we also gather process stats for this? like what we do in model queue: https://github.com/3drepo/3drepobouncer/blob/master/tools/bouncer_worker/src/queues/modelQueueHandler.js#L63

It'll be good to record them in elastic like we do for model imports so we can see resource utilisation etc

We might need a new index if it needs to differ from the import schema

tools/bouncer_worker/src/constants/messageTypes.js
```diff
@@ -0,0 +1,22 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ */
+
+module.exports = {
+	MSGTYPE_CLASH: 'clash',
```
carmenfan: Not a super important one, but it'd better follow our convention in the rest of the code base if we make this more like an enum object:



```suggestion
    messageTypes: {
         CLASH: 'clash',
         //etc
    }
	
```

So it's referenced as `messageTypes.CLASH` etc

bouncer/src/repo/core/model/bson/repo_node_revision.h
sebjf: Yes it was failing on gcc!

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_clearance.h
```diff
@@ -0,0 +1,68 @@
...
+#include "clash_pipelines.h"
+#include "repo/lib/datastructure/repo_uuid.h"
+#include "repo/lib/datastructure/repo_line.h"
+
+namespace repo {
+	namespace manipulator {
+		namespace modelutility {
+			namespace clash {
+
+				struct Pair
```
FJThiel: Is this actually used anywhere?

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_pipelines.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+		for (auto& obj : setA) {
+			if (intersection.find(obj.id) == intersection.end()) {
+				sets.a.push_back(&obj); // In A only
+			}
+		}
+	}
+}
+
+Pipeline::Pipeline(
```
FJThiel, line 260: Since this is empty, could this go into the header?

bouncer/src/repo/manipulator/modeloptimizer/bvh/single_ray_traverser.hpp
sebjf: Similar to ctime, gcc requires the header (not sure how msvc manages without it!)

bouncer/src/repo/core/handler/repo_database_handler_mongo.cpp
sebjf: @carmenfan this flag was added to make the unit tests run quicker iirc. @FJThiel can you say more about it? (and if there is another mechanism besides the flag - e.g. can we control the log level in the tests?)

bouncer/src/repo/core/model/bson/repo_node_mesh.h
sebjf: Yes that should be reference (as should the method above it as well!) Fixed.

bouncer/src/repo/lib/hopscotch_map/bhopscotch_map.h
sebjf: Yes, I realised the actual lib is snake case too so its just the folder. I've also moved it into lib with the json parser, as Felix spotted a couple of other places where it could be beneficial.

bouncer/src/repo/lib/datastructure/repo_range.cpp
sebjf: No, its just belt and braces if we wanted to use it later. Removed.

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_node_cache.cpp
sebjf: Same! removed.

bouncer/src/repo/manipulator/modelutility/spatialpartitioning/repo_spatial_partitioner_rdtree.cpp
sebjf: yes, I don't think you're missing anything, just compiler peculiarities 😄 not sure whether to be annoyed at gcc for being picky or msvc for letting it go so long!

bouncer/src/repo/manipulator/modelutility/repo_json_parser.h
sebjf: its already been moved once before so we might as well put it in the proper place once and for all 😄

bouncer/src/repo/manipulator/modelutility/clashdetection/geometry_exceptions.h
sebjf: Yes, the geometry tests are a dependency of the clash code, so this avoids the geometry code needing to bring in all the clash exceptions!

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_exceptions.h
sebjf: > Is there a reason why we don't inherit it?
We just didn't need the base class functionality, but agree & I have made it subclass `std::runtime_error` which inherits from `exception`.

> Perhaps we should rename them to Errors instead to avoid confusion?

I am in two minds, on one hand it is an exceptional case that prevents the engine from running, though it is not fatal to the process. On the other hand they are special in that they can copy themselves specifically to be put in the error report. However still, they use the same nested exception based error handling that we use in bouncer now, its just these particular exceptions can be routed to the results - which is how exception handlers are meant to work - save the process if possible, so they are consistent with best practice.

I decided to leave them as exceptions, but make them proper subclasses of the standard types!

bouncer/src/repo/manipulator/modelutility/clashdetection/geometry_tests.cpp
```diff
@@ -0,0 +1,837 @@
...
+		return T.b + (T.c - T.b) * (d4 - d3) / ((d4 - d3) + (d5 - d6));
+	}
+
+	auto denom = 1 / (va + vb + vc);
+	auto v = vb * denom;
+	auto w = vc * denom;
+	return T.a + ab * v + ac * w;
+}
+
+repo::lib::RepoLine geometry::closestPoints(const repo::lib::RepoVector3D64& p, const repo::lib::RepoTriangle& T)
```
FJThiel, line 91: Just sanity checking: this is called closestPointS (plural) because it returns a line with the closest point, which is the point in the triangle and p?

Just going from the name, I would have expected to get a collection of points e.g. from p to a collection of triangles.

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_clearance.cpp
sebjf: The lines are almost identical but the components of the vector differ in each one, and the nature of the vector type I felt would make any other implementation less clear overall. I have made the magic number a named define though and added a comment saying where the method came from and also some notes about what it should optimise for.

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_hard.cpp
sebjf: The big challenge with `run` is though the logic looks very similar, it actually differs in lots of tiny ways.

For example,

1. Clearance gets away with 1 cache object while hard uses 3
2. The type instantiations of the broadphase results vectors are different
3. The broadphase lambdas have slightly different signatures (due to (1))
4. The locking requirements between the cache entries are different (clash gets away with a shared lock while hard requires an exclusive lock on both objects). The narrowphase loop termination criteria are different too.
5. The narrowphase loops are different even during initialisation (e.g. one orders the vertices and the other doesn't have to)
6. The scheduling requirements are slightly different (hard requires parcelling the scheduled tasks to avoid contention, and clearance doesn't because it can use shared locks)

Some things I considered previously:

1. Turn the broadphase lambdas into virtual methods & move the duplicate inter & intra tests into the superclass. However, then the cache objects have to become subclass members, along with the broadphase functors. These are currently in anonymous namespaces within the files because they are only ever used in that one override of that one class. So it would be polluting more namespaces and spreading definitions over more files to save those 5-10 lines of duplicated code.

2. Make the narrowphase main loop a virtual method. However, then the multithreading primitives get spread out - currently 100% of the multithreading code exists within one function, on the same level, which is good for understanding how it operates.

3. Resolve some of this type complexity by making Pipeline take generic arguments. However this again means we have to place a lot of forward declarations around, and also means we need yet another parent class to be the non-generic Pipeline.

Overall, while it may be possible to de-duplicate by introducing some new superclasses I am not sure how much it is worth the reduction in clarity within each instance. Also, if we were to just about thread the needle, if we introduced a third mode it would probably blow it all up...

I did refactor createClashReport though.

Let me know if you want me to take another look!

bouncer/src/repo/manipulator/modelutility/repo_clash_detection_engine.cpp
```diff
@@ -0,0 +1,133 @@
...
+		// be indicated to the user via the report.
+
+		results.errors.push_back(std::move(e.clone()));
+	}
+
+	return results;
+}
+
+ClashDetectionEngine::ClashDetectionEngine(
+	std::shared_ptr<repo::core::handler::AbstractDatabaseHandler> handler)
```
FJThiel, line 68: Since this is empty, maybe it could be moved into the header?

bouncer/src/repo/manipulator/repo_manipulator.h
```diff
@@ -27,6 +27,7 @@
 #include "../core/model/bson/repo_node_transformation.h"
 #include "../core/model/bson/repo_node_mesh.h"
 #include "../core/model/collection/repo_scene.h"
+#include <repo/manipulator/modelutility/repo_clash_detection_config_fwd.h>
```
FJThiel, line 30: Why is the forward declaration here necessary? A least my MSVC compiler accepts it with the full header.

bouncer/src/repo/repo_controller.h
```diff
@@ -35,6 +35,7 @@
 #include "lib/datastructure/repo_structs.h"
 #include "lib/repo_config.h"
 #include "manipulator/modelconvertor/import/repo_model_import_config.h"
+#include <repo/manipulator/modelutility/repo_clash_detection_config_fwd.h>
```
FJThiel, line 38: Same as with the manipulater. My MSVC compiles without the forward declaration.

test/src/unit/repo/manipulator/modelutility/clashdetection/ut_repo_geometry.cpp
```diff
@@ -0,0 +1,1116 @@
...
+
+			EXPECT_THAT(vectors::Iterator(line), vectors::Intersects({ a.a, a.b, a.c, b.a, b.b, b.c }, { 1, 2 }, ClashGenerator::suggestTolerance({ a, b })));
+		}
+	}
+}
+
+TEST(Geometry, TrianglesUnitEE)
+{
+	// Tests that the triangle distance tests correctly identify triangles
+	// which are closest at an edge and a vertex.
```
FJThiel:
```suggestion
	// which are closest at their edges.
```

test/src/unit/repo/manipulator/modelutility/clashdetection/ut_repo_geometry.cpp
```diff
@@ -0,0 +1,1116 @@
...
+
+			EXPECT_THAT(e, Le(1.0));
+		}
+	}
+}
+
+TEST(Geometry, TrianglesUnitVF)
+{
+	// Tests that the triangle distance tests correctly identify triangles
+	// which are closest at an edge and a vertex.
```
FJThiel:
```suggestion
	// which are closest at a vertex and a face.
```

test/src/unit/repo/manipulator/modelutility/clashdetection/ut_repo_geometry.cpp
```diff
@@ -0,0 +1,1116 @@
...
+			EXPECT_THAT(e, Lt(1.0));
+		}
+	}
+}
+
+TEST(Geometry, TrianglesUnitFE)
+{
+	// Tests that the triangle distance tests correctly identify triangles
+	// where the edge of one penetrates the face of another.
+	// Note that the distance in this case should always be zero - if a
```
FJThiel:
```suggestion
	// Note that the distance in this case should always be zero - if
```

test/src/unit/repo/manipulator/modelutility/clashdetection/ut_repo_geometry.cpp
```diff
@@ -0,0 +1,1116 @@
...
+			auto b = random.direction();
+			auto n = random.direction();
+			auto c = repo::lib::RepoMatrix::rotation(n, random.number({ 1.57, 1.57 })) * b; // Rotation matrices are counterclockwise for positive angles
+
+			b = b * random.number(r1) + a;
+			c = c * random.number(r1) + a;
+
+			auto d = n + a;
+
+			//auto r = geometry::orient(a, b, c, d);
```
FJThiel: Leftover code or deliberate comment?

test/src/unit/repo/manipulator/modelutility/clashdetection/ut_repo_geometry.cpp
```diff
@@ -0,0 +1,1116 @@
...
+			// Check that the results are not changed by the winding order
+
+			clashGenerator.swapWindingOrder(a);
+			auto t3 = geometry::intersects(origin, direction, a);
+
+			EXPECT_THAT(t3, DoubleNear(t1, FLT_EPSILON));
+		}
+
+		{
+			// Two triangles with a shared ege, to test hitting the edge from inside and
```
FJThiel:
```suggestion
			// Two triangles with a shared edge, to test hitting the edge from inside and
```

test/src/unit/repo/manipulator/modelutility/clashdetection/ut_repo_geometry.cpp
```diff
@@ -0,0 +1,1116 @@
...
+			auto t2 = geometry::intersects(origin, -direction, a);
+
+			EXPECT_THAT(t2, Eq(INFINITY));
+
+			// Check that the results are not changed by the winding order
+
+			clashGenerator.swapWindingOrder(a);
+			auto t3 = geometry::intersects(origin, direction, a);
+
+			EXPECT_THAT(t3, DoubleNear(t1, FLT_EPSILON));
```
FJThiel, line 823: Shouldn't this use a double epsilon?

README.md
```diff
@@ -44,23 +44,19 @@ The following are optional dependencies depending on your configuration
...
-Compilation (Qt)
-------------
-See instructions at [https://github.com/3drepo/3drepobouncer/issues/9#issuecomment-135727927](https://github.com/3drepo/3drepobouncer/issues/9#issuecomment-135727927)
 
 Compilation (Windows)
 ------------
 Ensure Mongo CXX Driver, Boost and ASSIMP libraries are installed.
 
-The following instruction is for compiling a 64bit library using Visual Studio 12's tools. Change the pathing/cmake option appropriately if you are using another version of Visual Studio or compiling with a different compiler.
+The following instruction is for compiling a 64bit library using Visual Studio's tools. Change the pathing/cmake option appropriately if you are using another version of Visual Studio or compiling with a different compiler.
```
FJThiel:
```suggestion
The following instruction is for compiling a 64bit library using Visual Studio's tools. Change the pathing/cmake option appropriately if you are using a different compiler.
```

test/src/unit/repo/manipulator/modelutility/ut_repo_clash_detection.cpp
```diff
@@ -0,0 +1,4813 @@
...
+		// The tolerance for this test is set very high so we can make minor adjustments
+		// to the same problem to test the fingerprinting sensitivity.
+		// Small changes to the primitives (and so intersection location) should result
+		// in different fingerprints.
+
+		p.a.m = repo::lib::RepoMatrix::translate(repo::lib::RepoVector3D64(1, 0, 0)) * p.a.m;
+
+		auto run3 = pipeline.run({ p });
+
+		EXPECT_THAT(run1.clashes[0].fingerprint, Not(Eq(run3.clashes[0].fingerprint)));
```
FJThiel: Why Not(Eq(...)) and not Ne(...)?

(I only commented it once here, but there are quite a few of that here in the file)

test/src/unit/repo_test_matchers.h
```diff
@@ -195,6 +199,160 @@ namespace testing {
...
+		if (abs(a.z - b.z) > tolerance) { *listener << "Vector z exceeds tolerance."; return false; }
+		return true;
+	}
+
+	MATCHER_P2(UnorderedVectorsAre, elements, tolerance, "")
+	{
+		std::vector<repo::lib::RepoVector3D64> vectors(elements);
+		for (auto& r : arg) {
+			for (auto i = 0; i < vectors.size(); i++) {
+				if (compareVectors(r, vectors[i], tolerance, result_listener)) {
```
FJThiel, line 216: Accessing a collection by index with an i capped by the size _while_ deleting elements... is that safe? Will the last elements even be checked as .size() would decrease with each erasure, while i increases with each step?

test/src/unit/repo_test_utils.cpp
```diff
@@ -37,9 +37,9 @@ repo::RepoController::RepoToken* testing::initController(repo::RepoController* c
 repo::lib::RepoVector3D testing::makeRepoVector()
```
FJThiel: maybe makeRandomRepoVector() to make it clear that it will not just be a vector, but one initialised with random values.

Edit: I just saw that we have a the random generator which also does vectors. Could we use that here?

test/src/unit/repo_test_random_generator.cpp
```diff
@@ -0,0 +1,209 @@
...
+	return d(gen);
+}
+
+double RepoRandomGenerator::number(double upper)
+{
+	std::uniform_real_distribution<> d(0, upper);
+	return d(gen);
+}
+
+int64_t RepoRandomGenerator::range(int64_t lower, int64_t upper)
```
FJThiel: The method name sounds to me like it would return a random range, but it returns a single int.

FJThiel (changes requested):

bouncer/src/repo/core/handler/repo_database_handler_mongo.cpp
FJThiel: This was for the multithreading soak tests. The purpose is actually the other way around i.e. we want it on release and not in debugging. On release, we do print out what indices are created. Whether this is necessary or not, I can't judge, but it always was this way since I joined, so I would hope someone saw a value in it.

Now, the soak tests for the database handler creates thousands of these indices across multiple threads while simultaneously performing other operations to ensure that they don't collide with each other. But printing out thousands of these lines slows them down quite significantly. I will see whether the log level in gtest can be altered, but, at least in general, we want more output in tests not less, hence I added that flag so other output can still make it through.

Or we just ditch that output line on index creation altogether, if it has no real value.

bouncer/src/repo/manipulator/modelutility/repo_clash_detection_config.cpp
sebjf: Yes to both! I have made it so the SAX based parser will throw if it encounters the wrong token, and I have also added an explicit validation step in the client. In both cases bouncer ends up wrapping the inner exception in a new RepoInvalidConfigException exception & returns the appropriate exit code.

The validation happens explicitly in the client, because some technically invalid configs should pass through the engine without crashing (and are tested as such).

tools/bouncer_worker/src/queues/clashQueueHandler.js
sebjf: @carmenfan, we can just omit the owner (or add a constant value) and then the schema is the same?

```
		const processInformation = Utils.gatherProcessInformation(
			'', // headless clashes have no owner
			project,
			teamspace,
			logLabel.label, // queue
			config.repoLicense,
			rid.toString(),
		);
```

carmenfan: @sebjf  yes that's fine we did something similar when we had unity processing

bouncer/src/repo/core/handler/repo_database_handler_mongo.cpp
sebjf: @FJThiel, if its not too much trouble, we should make the soak tests turn logging down or off. This will have to be done on a per-test basis. It will also have to be done on the logger object itself as the stdout/stderr streams don't have that level of control.

bouncer/src/repo/core/handler/database/repo_query.h
sebjf: Unfortunately not - the reason being is that while `std::string` does have an implicit conversion from `const char*`, if we rely on that and make them all `std::string`, the compiler in this case reverts to the templated fallback `Eq(const std::string& field, const T& value)`, in which it tries to interpret the vector for oneOf, anyOf etc as a single value. (Not sure what compiler rule results in this, but basically all invocations were we want to match one of the prior overloads, the types need to match exactly, which means we need const char* templates to do things like:
```
Eq("myConstantField", { vector, of, stuff });
```

bouncer/src/repo/lib/datastructure/repo_matrix.cpp
sebjf: @FJThiel, we do yes; to expand... 

All our matrices are row major in the sense that we always pre-multiply (i.e. it is always Matrix * Vector, and never Vector * Matrix). The flag exists on the constructor to say whether the coefficients are laid out row-wise or column-wise, when loading the row-major matrix. This is actually distinct from whether the transform matrix is row or column major (see the glMatrix library for an (annoying) example of this!).

I don't think we have any importers atm which provide matrices that are laid out column-wise, and we always store them row-wise, so this flag is probably never false in the current codebase outside the unit tests.

bouncer/src/repo/lib/datastructure/repo_matrix.cpp
sebjf: Whilst I am not sure I like it 😆 it will never fail incomplete - this only triggers once the matrix has been filled up. It could also be written...

```
	for (int i = 0; i < 16; ++i)
	{
		data[i] = (T)mat[i];
	}
```

..and in that way is no different really to the constructor that takes a pointer.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_dgn.cpp
sebjf: Do you mean the check in data_processor_rvt? The reason rvt has that check is that it processes layers sort of like a stack, and not all 'stack frames' will have their own context. (i.e. it is not a safety check - its part of the logic). The dgn importer creates the context right above so doesn't need it!

bouncer/src/repo/manipulator/modelutility/clashdetection/bvh_operators.cpp
sebjf: it means distance² (it doesn't take the sqrt of the return value, meaning its faster if we only need to compare like for like and don't care about the absolute distance in true terms).

bouncer/src/repo/manipulator/modelutility/clashdetection/clash_pipelines.cpp
sebjf: I would say it is not really empty. An empty implementation could be deleted (and so would be deleted instead of moved into the header).

Initialising handler and config are genuine operations, such that if we removed that snippet, the class would break.

Therefore it is an implementation and so belongs in the cpp file!

bouncer/src/repo/manipulator/modelutility/clashdetection/geometry_tests.cpp
sebjf: >  because it returns a line with the closest point, which is the point in the triangle and p?

Correct, the convention is closetPoints return lines and closestPoint returns an individual point.

bouncer/src/repo/manipulator/modelutility/repo_clash_detection_engine.cpp
sebjf: Again it is not really empty it just uses a language shortcut to hide the assignment!

bouncer/src/repo/manipulator/repo_manipulator.h
sebjf: Passing the full header is OK, but it is quite a big header. For files that just need references, it improves compiles time if we don't have to include the full header.

bouncer/src/repo/repo_controller.h
sebjf: MSVC might compile without a declaration but Gcc won't!

test/src/unit/repo/manipulator/modelutility/clashdetection/ut_repo_geometry.cpp
sebjf: FLT_EPSILON is a convenient threshold, while DoubleNear is based on the type. We can't use DBL_EPSILON because the tests themselves have way more quantisation error than that!

test/src/unit/repo_test_matchers.h
sebjf: Indeed if we had an erase followed by another comparison it would not be safe (it would not cause an access violation, because size() would be re-evaluated, but it would miss vectors). 

However that is not what is happening here; the outer loop iterates over `arg`. The inner loop checks if `arg` matches any remaining element of `vectors`, and if so deletes it then breaks out of the inner loop. So the iteration over `vectors` always starts from the beginning.

test/src/unit/repo_test_utils.cpp
sebjf: I've updated the name to be consistent with the other utils members. 

We do have a new and improved random class, however it is not so straightforward to swap it as this method is used in legacy tests, and is actually sensitive to the global state (i.e. resetRand). Eventually every time those tests are touched they should be upgraded to use the new random object, but that is out of scope of this issue!

bouncer/src/repo/lib/datastructure/repo_bounds.cpp
```diff
@@ -67,4 +93,83 @@ const RepoVector3D64& RepoBounds::max() const
...
```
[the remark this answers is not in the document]
sebjf: (Same as before, the initialiser calls the constructor which is robust to this.)

bouncer/src/repo/manipulator/modelutility/repo_clash_detection_config.cpp
```diff
@@ -0,0 +1,251 @@
...
+	config = {};
+	std::ifstream fileStream(jsonFilePath, std::ios::in | std::ios::binary);
+	std::ostringstream contentStream;
+	contentStream << fileStream.rdbuf();
+	auto content = contentStream.str();
+	try{
+		ClashConfigParser::ParseJson(content.data(), config);
+	}
+	catch (const JsonParser::JsonParsingException& e) {
+		std::throw_with_nested(repo::lib::RepoInvalidConfigException("Failed to parse clash detection config"));
```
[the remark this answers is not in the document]
sebjf: Throw with nested maintains the original JsonParsingException

test/src/unit/repo/manipulator/modelutility/clashdetection/ut_repo_bvh_operators.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+	clashGenerator.downcastVertices = false;
+
+	auto hyperplanes = std::vector<repo::lib::RepoVector3D64>{
+		repo::lib::RepoVector3D64(1, 0, 0),
+		repo::lib::RepoVector3D64(0, 1, 0),
+		repo::lib::RepoVector3D64(0, 0, 1)
+	};
+
+	for (int itr = 0; itr < 100000; ++itr) {
```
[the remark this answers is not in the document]
sebjf: (Same as before, future task)

tools/bouncer_worker/src/lib/config.js
```diff
@@ -101,6 +101,18 @@ const schema = object({
 	elastic,
 });
 
+/* Throws if the directory doesn't exist or write access is not available */
+const checkDirectory = (directory) => {
+	if (!fs.existsSync(directory)) {
+		throw `Shared directory does not exist: ${directory}`;
+	}
+	try {
+		fs.accessSync(directory, fs.constants.R_OK);
```
carmenfan: The comment says we're checking for write access, but the code seems to be checking for read? should it be this instead?
```` js
fs.accessSync(directory, fs.constants.R_OK | fs.constants.W_OK);
````

carmenfan (dismissed):
@sebjf mostly looks fine, just one question around fs.access params!

tools/bouncer_worker/src/lib/config.js
sebjf: 😮 yes it should! Fixed!

Issue #782, opened 2025-07-02
Create a bouncer_worker utility to upgrade all unity bundles


sebjf:
Product Ticket: https://github.com/3drepo/3D-Repo-Product-Team/issues/549

### Description

To support future features, we need to [decommission remaining Unity AssetBundles](https://github.com/3drepo/3D-Repo-Product-Team/issues/549).

There are two parts to this, one is to actually generate RepoBundles for an existing scene collection, so that the instance UUIDs (which are referenced by tickets, views, etc) remain unchanged. Then, we need to ensure that any assets cached for that revision do not interfere with loading the new one, which is primarily a .io task.

This ticket is for the asset rebuild functionality.

Bouncer has a genStash command which can be used to rebuild the RepoBundles (and/or tree) for a given revision or revisions.

The first thing to do is to make sure this still works OK after all the recent changes to bouncer.

Then we need a way to invoke it to rebuild the necessary scenes.

Next, we need a way to signal to the frontend that it should invalidate the necessary cache entries (i.e. the supermeshes.json.mpc response), because the underlying assets have changed.

Finally, a nice to have would be that it also cleans up previous assets to save space (deleting the ref node would be enough then the zombie script will clean up the disk), though this could also be made the responsibility of a migration script (probably safest...)

sebjf:
Closing as the migration is now done. For future migrations, we have recorded divergences in the schema that must be addressed here: https://github.com/3drepo/3D-Repo-Product-Team/issues/1020

PR #832, opened 2026-03-20
ISSUE #782 Unity asset bundles migration utility

sebjf:
#### Description
This PR exists in order to create an image of #782. 782 is a special branch that embeds a utility in bouncer_worker to migrate all revisions that only have Unity AssetBundles to RepoBundles.

THIS PR SHOULD NEVER ACTUALLY BE MERGED!

The branch includes the modified worker as well as minor tweaks to bouncer itself to get the desired behaviour. That is, the bouncer_worker and bouncer from this branch should always be built and run together.

That desired behaviour is:

All Unity-only revisions - defined as revisions that have a non-incomplete entry in `.history`, but no matching entry in `.stash.repobundles` - should have RepoBundles generated in place. In-place means that:

1. Groups and Saved Views will continue to work.
2. Void status of revisions are preserved. Voided revisions **are** migrated, so can be loaded if the status is removed.
3. Coordinates remain consistent (GIS, Annotations, etc register in the same place).
4. Timestamp reflects the original import time, not the migration time.
5. Author reflects the original uploader, not the account of the migration process.

To start the migration, deploy the image, but with the command to start bouncer changed to `yarn run-migration` instead of `yarn run-worker`.

Passing a teamspace (e.g. `yarn run-migration sebjf`) will migrate only that teamspace. Only one teamspace can be passed at a time. If no teamspace is specified, all teamspaces will be processed.

The process will terminate when all revisions have been migrated.

If a migration is performed for a revision, that revision is done & will not be picked up if the process is run again, so the process can be executed as many times as needed - however - instances *must not run in parallel* as the utility gets the list of revisions to process in-advance.

If the process fails for whatever reason, it *is* possible for a revision to be left in the `incomplete` state. In this case, running the utility again will *not* recover the revision because it ignores incomplete revisions (assuming they failed on import, and so the .scene collection is incomplete). If this happens, the revision must be identified from the log and the `incomplete` flag removed manually before trying again.


carmenfan (changes requested):

carmenfan (changes requested):

sebjf:
Closing because we are now done with the migration.

PR #837, opened 2026-03-27
Issue #836 added duplication of material nodes if it exceeds the usag…

### LINKED ISSUE #836: Allow materials to be shared by more mesh nodes
### Description
Product issue: [#717](https://github.com/3drepo/3D-Repo-Product-Team/issues/717)
(Discovered during work on that product ticket)

While investigating large models in relation of the tree, it was found that one with about 5 million nodes does not import. The source of the crash was found to be that the document updating the parent field of a material node exceeded the allowed document size of the mongo db. This happened, because that particular material was used by about 708 thousand mesh nodes. Given that the model itself had some 1.7 million mesh nodes, that is not unexpected behaviour and should be fixed.

### Goals

- [x] The user should be able to upload meshes where 700k or more meshes use the same material

### Tasks

- [x] Implement a mechanic that duplicates a material node if a certain threshold is crossed so that a material node cannot have more parents than the document size allows.

FJThiel:
This fixes #836

#### Description

These changes will duplicate material nodes if they have been referenced more than 500k times. This allows for more than 500k mesh nodes to use the same material properties without the material node exceeding the maximum document size in the database.

This will allow for the import of larger models.


sebjf (commented):
Brief question about the logic!

sebjf (approved):

PR #854, opened 2026-04-24
#855 initial deploy workflow file

### LINKED ISSUE #855: Bouncer deployment via workflow
WE now have the building of bouncer via a workflow in bouncer repo. Extend this to deploy bouncer on staging too like we have for io

nkuberaf:
This fixes #855 

#### Description
#855  create deploy workflow for bouncer

The flow is now:             
  - workflow_run fires automatically after "Build and Push Bouncer" succeeds on the staging branch, always deploys staging image.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -97,7 +97,12 @@ jobs:
             # Allow /deploy-bouncer <tag> to specify an explicit tag; fall back to PR branch
             EXPLICIT_TAG=$(echo "$COMMENT_BODY" | awk '{print $2}')
             if [ -n "$EXPLICIT_TAG" ] && [ "$EXPLICIT_TAG" != "null" ]; then
-              IMAGE_TAG="$EXPLICIT_TAG"
+              if [[ "$EXPLICIT_TAG" =~ ^[A-Za-z0-9._-]+$ ]]; then
+                IMAGE_TAG="$EXPLICIT_TAG"
+              else
+                echo "Invalid explicit image tag: only letters, numbers, dot, underscore, and hyphen are allowed"
+                exit 1
+              fi
             else
               ISSUE_NUMBER=$(jq -r .issue.number < "$GITHUB_EVENT_PATH")
               IMAGE_TAG=$(curl -s \
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -116,8 +116,12 @@ jobs:
             fi
           fi
 
-          echo "image_tag=$IMAGE_TAG" >> $GITHUB_OUTPUT
+          if [[ "$IMAGE_TAG" == *$'\n'* || "$IMAGE_TAG" == *$'\r'* ]]; then
+            echo "Resolved image tag contains unsupported newline characters"
+            exit 1
+          fi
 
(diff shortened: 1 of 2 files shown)


carmenfan (changes requested):
As discussed in meeting, let's simplify it for the use case requested (auto deployment on new staging commits) and worry about the rest later.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,73 @@
...
+      # ---------------------------------------------------------------
+      - name: Deploy Helm chart
+        run: |
+          helm upgrade --install \
+            --version 5.22.2 \
+            www-bouncer \
+            oci://${{ secrets.REGISTRY_LOGIN_SERVER }}/helm/bouncer \
+            --namespace default \
+            -f deployments/asite-stg/helm/bouncer/values.yaml \
+            --set image.tag=staging \
```
carmenfan: Is it possible to make this get the info from the branch instead of hard coding the name?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,73 @@
...
+          echo "${{ secrets.REGISTRY_TOKEN }}" | helm registry login "${{ secrets.REGISTRY_LOGIN_SERVER }}" \
+            -u "${{ secrets.REGISTRY_USERNAME }}" --password-stdin
+
+      # ---------------------------------------------------------------
+      # STEP 5: Deploy Helm chart
+      # ---------------------------------------------------------------
+      - name: Deploy Helm chart
+        run: |
+          helm upgrade --install \
+            --version 5.22.2 \
```
carmenfan: Can we make this a variable up top so whenever we want to update the helm chart we just change the values on the top of this file (like declaring it as an env var near the beginning) instead of having to fish around for it.

Similar to what is done here: https://github.com/3drepo/3drepobouncer/blob/master/.github/workflows/runUnitAndIntegrationTests.yml#L12

Possibily the same about image tag (but with branch name extracted from the action, ideally)

carmenfan (changes requested):

.github/workflows/deploy.yml
carmenfan: Also, I believe you would have to change the build and push action to trigger on every commit in staging?

On top of that, what if build and push were triggered by a tag - if I tagged a commit on staging, do I end up triggering the script twice? (or do tag-triggered scripts not have the branch info thus it won't trigger the deploy script?)

nkuberaf: i think build workflow (dockerBuildAndPush.yml) needs a push: branches: [staging] trigger added i will look into that.

nkuberaf: A tag push triggers the build with e.g. GITHUB_REF = refs/tags/v5.22.2, so head_branch in the workflow_run event is v5.22.2 not staging. The branches: [staging] filter in deploy.yml won't match, so i think there will be no double trigger

.github/workflows/deploy.yml
nkuberaf: added

.github/workflows/deploy.yml
nkuberaf: updated buildandpush

carmenfan (changes requested):
Can you review copilot feedback?

carmenfan (approved):

PR #861, opened 2026-05-01
Issue 859 - support v5 configs

### LINKED ISSUE #859: Bouncer messages and config files should use new terminology
Product Ticket: https://github.com/3drepo/3D-Repo-Product-Team/issues/639

### Description

Update bouncer and bouncer_worker, such that..

Database -> Teamspace
Project -> Container or Drawing

This ticket will update bouncer client to use both v4 and v5 conventions, so the unit tests will not change. bouncer_worker will only support v5 conventions.

### Goals

- [x] All existing test configs still work
- [x] The backend can use v5 conventions only

### Tasks

### Related Resources

MockUp : `link to mockup`

sebjf:
This fixes #859

#### Description
This makes minor changes so that bouncer_worker operates exclusively with v5 conventions, and bouncer itself can use v4 or v5 conventions.

Only the model and drawing functions have been updated in bouncer, as genStash operates exclusively with command line arguments, and federations are no longer supported.

Clash detection does not need any changes as this already/only works with the new conventions.

### DIFF
--- a/client/src/functions.cpp
+++ b/client/src/functions.cpp
@@ -404,8 +404,18 @@ int32_t importFileAndCommit(
 			tag = jsonTree.get<std::string>("tag", "");
 			desc = jsonTree.get<std::string>("desc", "");
 
-			config.databaseName = jsonTree.get<std::string>("database", "");
-			config.projectName = jsonTree.get<std::string>("project", "");
+			// Client should be backwards compatible with v4 (database & project) and v5
+			// (teamspace & container) terminology. Teamspace/database are the terms that
+			// do not conflict, so we use them to distinguish which version is being used.
+
+			config.databaseName = jsonTree.get<std::string>("teamspace", "");
+			config.projectName = jsonTree.get<std::string>("container", "");
+
+			if (config.databaseName.empty()) {
+				config.databaseName = jsonTree.get<std::string>("database", "");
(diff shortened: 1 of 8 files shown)


client/src/functions.cpp
```diff
@@ -417,8 +417,18 @@ int32_t importFileAndCommit(
...
-			config.projectName = jsonTree.get<std::string>("project", "");
+			// Client should be backwards compatible with v4 (database & project) and v5
+			// (teamspace & container) terminology. Teamspace/database are the terms that
+			// do not conflict, so we use them to distinguish which version is being used.
+
+			config.databaseName = jsonTree.get<std::string>("teamspace", "");
+			config.projectName = jsonTree.get<std::string>("container", "");
+
+			if (config.databaseName.empty()) {
+				config.databaseName = jsonTree.get<std::string>("database", "");
```
FJThiel, line 428: Given that the "new" terminology is now teamspace and container, should our variables here reflect that e.g. config.teamspace and config.container?

sebjf: I started to do this but the number of changes is.. large. I think it would be better to do those as a separate maintenance issue where that is literally the only thing being done. Esp because any form of semi-automation is going to have trouble with the word 'database'...

FJThiel (approved):

PR #862, opened 2026-05-01
Issue 842 - Support splitting by floors

### LINKED ISSUE #842: Build supermeshes by Floor and tag the Bundles accordingly
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/943

Platforms supporting offline mode (e.g. Field), would like to only download assets associated with specific floors.

This could be achieved by first grouping elements by Floor before Supermeshing, and recording those floors along with the metadata of the RepoBundles associated with those floors.

The platform can then decide which assets it makes available to the viewer for streaming (mechanism tbc).

This will only be supported for formats with complete BIM data.

### Goals

- [x] When toggled on, Supermeshes are created such that meshes belonging to one floor are always grouped together.
- [x] Floor metadata exists even when the splitting is off in the config

### Tasks
- [x] Look through a bunch of vertical models and identify which metadata, if any, can be reliably used to distinguish floors
- [x] See if that metadata can be identified on input or if we require a smart group
- [x] Make some test files that represent all the level constraints (3 days)
- [x] Get config passed through to the correct places (1 day)
- [x] Add metadata params to Rvt importer (1 day)
- [x] Add metadata params to Ifc importer (1 day)
- [x] Create unit tests for metadata params (1 day)
- [x] Create unit tests for Ifc metadata params
- [x] Add support for Nwds (when hosting Ifc or Rvt files)
- [x] Create unit tests for Nwds
- [x] Add stage to multipart optimiser to group meshes (5 day)
- [x] Ensure optimiser is respecting groups and add unit tests (5 day)
- [x] Add grouping info to assets response (2 day)
- [x] Validation on test models (2 day)

### Related Resources

MockUp : `link to mockup`

sebjf:
This fixes #842

#### Description
This PR adds support for splitting up mesh nodes into bundles based on the tree, as well as resource ids (meshnode grouping members). Additionally, when split by branch, the bundles are tagged.

The purpose is to allow downloading just some floors in offline mode, though for now the feature is on by default meaning models can be used offline post-hoc.

The way it works is that when a transformation node that represents the tip of a branch corresponding to a floor/level is encountered, a magic metadata parameter is added to it. Currently this is done for IFC, RVT and NWD files.

Later on, during optimisation, mesh nodes are grouped based on whether they are descendants of one of these nodes. Nodes are split by both branch and resource groups.

In the future, this stage could be extended to support grouping by other metadata queries (e.g. to cut up long linear infrastructure models).

To support this, the `clusterAndSupermesh` method has been modified so it that performs the grouping, rather than jobs being built around mesh node groups. Additionally, the grouping member has been added to the SupermeshNode type (for output to the assets document), and the optimiser has been updated to pass this on.

All grouping operations take place in `clusterAndSupermesh`, allowing the grouping behaviour to updated easily in the future.

Finally, a few quality of life improvements are made: the combine_hash function has been moved into a header, the repo queries have been updated to avoid the const char* converting to bool issue happening, and the streaming mesh node fails fast  & early with a nested exception if its contents are accessed before it is loaded.

### DIFF
--- a/bouncer/src/repo/core/model/bson/repo_bson_assets.h
+++ b/bouncer/src/repo/core/model/bson/repo_bson_assets.h
@@ -41,6 +41,14 @@ namespace repo {
 				size_t primitive;
 				repo::lib::RepoVector3D min;
 				repo::lib::RepoVector3D max;
+				std::vector<std::string> groups;
+
+				void addGrouping(const std::string& group)
+				{
+					if (!group.empty()) {
+						groups.push_back(group);
+					}
+				}
 			};
 
 			class REPO_API_EXPORT RepoAssets
--- a/bouncer/src/repo/core/model/bson/repo_bson_assets.h
+++ b/bouncer/src/repo/core/model/bson/repo_bson_assets.h
@@ -43,6 +43,24 @@ namespace repo {
 				repo::lib::RepoVector3D max;
 				std::vector<std::string> groups;
 
+				bool operator==(const RepoSupermeshMetadata& other) const
+				{
+					return numVertices == other.numVertices &&
+						numFaces == other.numFaces &&
+						numUVChannels == other.numUVChannels &&
+						numSubmeshes == other.numSubmeshes &&
+						numComponents == other.numComponents &&
+						primitive == other.primitive &&
+						min == other.min &&
+						max == other.max &&
+						groups == other.groups;
+				}
+
+				bool operator!=(const RepoSupermeshMetadata& other) const
+				{
+					return !(*this == other);
+				}
+
 				void addGrouping(const std::string& group)
 				{
 					if (!group.empty()) {
--- a/bouncer/src/repo/core/model/bson/repo_bson_assets.h
+++ b/bouncer/src/repo/core/model/bson/repo_bson_assets.h
@@ -43,18 +43,7 @@ namespace repo {
 				repo::lib::RepoVector3D max;
 				std::vector<std::string> groups;
 
-				bool operator==(const RepoSupermeshMetadata& other) const
-				{
-					return numVertices == other.numVertices &&
-						numFaces == other.numFaces &&
-						numUVChannels == other.numUVChannels &&
-						numSubmeshes == other.numSubmeshes &&
-						numComponents == other.numComponents &&
-						primitive == other.primitive &&
-						min == other.min &&
-						max == other.max &&
-						groups == other.groups;
-				}
+				bool operator==(const RepoSupermeshMetadata& other) const;
 
 				bool operator!=(const RepoSupermeshMetadata& other) const
 				{
--- a/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_nwd.cpp
+++ b/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_nwd.cpp
@@ -24,6 +24,7 @@
 #include "repo/manipulator/modelconvertor/import/odaHelper/repo_mesh_builder.h"
 #include "repo/core/model/bson/repo_node_transformation.h"
 #include "repo/core/model/bson/repo_bson_factory.h"
+#include "repo/lib/datastructure/repo_variant_utils.h"
 
 #include <repo_log.h>
 
@@ -926,6 +927,14 @@ OdResult traverseSceneGraph(OdNwModelItemPtr pNode, RepoNwTraversalContext conte
 
 			context.parentNode = context.sceneBuilder->addNode(RepoBSONFactory::makeTransformationNode({}, levelName, { context.parentNode->getSharedID() }));
 
+			// To match the plug-in, and ensure metadata ends up in the right place for
+			// the benefit of smart groups, node properties are overridden with their
+			// parent's metadata
+			std::unordered_map<std::string, repo::lib::RepoVariant> merged = metadata;
+			for (auto p : context.parentMetadata) {
+				merged[p.first] = p.second;
+			}
+
 			// GetIcon distinguishes the type of node. This corresponds to the icon seen in
 			// the Selection Tree View in Navisworks.
 			// https://docs.opendesign.com/bimnv/OdNwModelItem__getIcon@const.html
@@ -935,14 +944,27 @@ OdResult traverseSceneGraph(OdNwModelItemPtr pNode, RepoNwTraversalContext conte
 			if (pNode->getIcon() == NwModelItemIcon::LAYER)
 			{
 				context.layer = pNode;
-			}
 
-			// To match the plug-in, and ensure metadata ends up in the right place for
-			// the benefit of smart groups, node properties are overridden with their
-			// parent's metadata
-			std::unordered_map<std::string, repo::lib::RepoVariant> merged = metadata;
-			for (auto p : context.parentMetadata) {
-				merged[p.first] = p.second;
+				// For layers, check if we need to add the magic floor metadata. This applies
+				// to Rvt and Ifc partitions - both of which will create Layer's for floors.
+				//
+				// This magic parameter should be applied directly to the node's - it should
+				// not be inherited like the other parameters above. Hence we check the map
+				// extracted from the node (metadata) but write to map that is written to the
+				// db (merged).
+				{
+					auto it = metadata.find("Item::Internal Type"); // For Rvt files, look for the LcRevitLayer type
+					if (it != metadata.end() && boost::apply_visitor(repo::lib::StringConversionVisitor(), it->second) == "LcRevitLayer") {
+						merged[REPO_METADATA_GROUPING_FLOOR] = levelName;
+					}
+				}
+
+				{
+					auto it = metadata.find("Element::IfcClass"); // For Ifc files, look for the IfcBuildingStorey class
+					if (it != metadata.end() && boost::apply_visitor(repo::lib::StringConversionVisitor(), it->second) == "IfcBuildingStorey") {
+						merged[REPO_METADATA_GROUPING_FLOOR] = levelName;
+					}
+				}
 			}
 
 			context.sceneBuilder->addNode(RepoBSONFactory::makeMetaDataNode(merged, context.parentNode->getName(), { context.parentNode->getSharedID() }));
--- a/bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.cpp
+++ b/bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.cpp
@@ -56,7 +56,8 @@ MultipartOptimizer::MultipartOptimizer(
 	repo::manipulator::modelconvertor::AbstractModelExport* exporter
 ):
 	handler(handler),
-	exporter(exporter)
+	exporter(exporter),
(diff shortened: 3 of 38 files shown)


bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.cpp
```diff
@@ -181,55 +134,106 @@ std::unordered_map<repo::lib::RepoUUID, repo::lib::RepoMatrix, repo::lib::RepoUU
...
-				auto parentId = bson.getUUIDFieldArray(REPO_NODE_LABEL_PARENTS)[0];
-
-				if (childNodeMap.contains(parentId)) {
-					childNodeMap.at(parentId).push_back(bson);
-				}
-				else {
-					auto children = std::vector<repo::core::model::RepoBSON>{ bson };
-					childNodeMap.insert({ parentId, children });
-				}
+	repo::core::model::RepoBSON rootNode;
```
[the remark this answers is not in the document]
sebjf: Changed findAllByCriteria so it can no longer return nullptr

bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.cpp
```diff
@@ -181,55 +134,106 @@ std::unordered_map<repo::lib::RepoUUID, repo::lib::RepoMatrix, repo::lib::RepoUU
...
+	projection.includeField(REPO_NODE_LABEL_SHARED_ID);
+	projection.includeField(REPO_NODE_LABEL_PARENTS);
+	projection.includeField(REPO_NODE_LABEL_METADATA);
+
+	auto sceneCollection = collection + "." + REPO_COLLECTION_SCENE;
+	auto cursor = handler->findCursorByCriteria(database, sceneCollection, filter, projection);
+
+	std::unordered_map<std::string, BranchGroup*> branchGroupsByTag;
+
+	for (auto bson : (*cursor)) {
```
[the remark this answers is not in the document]
sebjf: Changed findAllByCriteria so it can no longer return nullptr

bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.cpp
```diff
@@ -433,64 +388,87 @@ void MultipartOptimizer::clusterAndSupermesh(
...
+		group,
+		std::vector<repo::core::model::StreamingMeshNode>,
+		group_hash,
+		group_equal
+	> groups;
+
+	// iterate cursor and pack outcomes in lightweight mesh node structure
+	for (auto bson : (*cursor)) {
+		repo::core::model::StreamingMeshNode node(bson);
+		auto transform = transformMap.at(node.getParent());
```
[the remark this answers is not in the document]
sebjf: This is not undefined behaviour: [`at`](https://en.cppreference.com/cpp/container/unordered_map/at) will throw std::out_of_range which is the intended fail-fast behaviour.

bouncer/src/repo/core/model/bson/repo_bson_assets.h
```diff
@@ -41,6 +41,14 @@ namespace repo {
 				size_t primitive;
 				repo::lib::RepoVector3D min;
 				repo::lib::RepoVector3D max;
+				std::vector<std::string> groups;
+
```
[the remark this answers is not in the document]
sebjf: Made suggested commit and removed the local operator in the unit tests

bouncer/src/repo/core/model/bson/repo_bson_assets.h
```diff
@@ -41,6 +41,32 @@ namespace repo {
...
+				{
+					return numVertices == other.numVertices &&
+						numFaces == other.numFaces &&
+						numUVChannels == other.numUVChannels &&
+						numSubmeshes == other.numSubmeshes &&
+						numComponents == other.numComponents &&
+						primitive == other.primitive &&
+						min == other.min &&
+						max == other.max &&
+						groups == other.groups;
```
FJThiel: If I understand the comparison of vectors via == correctly, this will only work if the groups are in the same order in both vectors. Is that guaranteed?

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_nwd.cpp
```diff
@@ -935,14 +944,27 @@ OdResult traverseSceneGraph(OdNwModelItemPtr pNode, RepoNwTraversalContext conte
...
-				merged[p.first] = p.second;
+				// For layers, check if we need to add the magic floor metadata. This applies
+				// to Rvt and Ifc partitions - both of which will create Layer's for floors.
+				//
+				// This magic parameter should be applied directly to the node's - it should
+				// not be inherited like the other parameters above. Hence we check the map
+				// extracted from the node (metadata) but write to map that is written to the
+				// db (merged).
+				{
+					auto it = metadata.find("Item::Internal Type"); // For Rvt files, look for the LcRevitLayer type
```
FJThiel, line 956: Rvt files and IFC classes are mentioned here multiple times. However, this is the data processor for NWD files. Is this for metadata in files that were originally Rvt or IFC and then converted to NWD?

bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.h
```diff
@@ -81,13 +87,33 @@ namespace repo {
...
+					std::string name; // The tag that gets associated with any supermeshes
+				};
+
+				/*
+				* Contains the baked transformation information, i.e. the final conditions
+				* of the Transformation just above a MeshNode.
+				*/
+				struct TransformInfo {
+					repo::lib::RepoMatrix matrix;
+					BranchGroup* branch = nullptr;
```
FJThiel, line 101: Maybe a smart pointer here to ensure that it will always be cleaned up properly?

bouncer/src/repo/manipulator/modelutility/repo_scene_manager.cpp
```diff
@@ -251,14 +254,15 @@ bool SceneManager::generateWebViewBuffers(
 			return false;
 		}
 
-		repo::manipulator::modeloptimizer::MultipartOptimizer mpOpt;
-		return mpOpt.processScene(
+		repo::manipulator::modeloptimizer::MultipartOptimizer mpOpt(handler, exporter.get());
+		mpOpt.splitByFloor = config.splitByFloor;
```
FJThiel: Why not set this through a constructor or accessor of the mpOpt?

bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.h
```diff
@@ -44,21 +44,27 @@ namespace repo {
...
+				void processScene(
 					std::string database,
 					std::string collection,
-					repo::lib::RepoUUID revId,
-					repo::core::handler::AbstractDatabaseHandler *handler,
-					repo::manipulator::modelconvertor::AbstractModelExport *exporter
+					repo::lib::RepoUUID revId
 				);
 
+				bool splitByFloor;
```
FJThiel: Does this need to be public?

bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.cpp
```diff
@@ -44,130 +47,80 @@ static const size_t REPO_MODEL_LOW_CLUSTERING_RATIO = 0.2f;
...
-			{
-				std::string description = "Grouping: " + grouping + ", Textured " + texId.toString() + " , Primitive 3, No Normals";
-				jobs.push_back(createTexturedJob(description, revId, 3, grouping, false, texId));
-			}
-			{
-				std::string description = "Grouping: " + grouping + ", Textured " + texId.toString() + " , Primitive 3, Normals";
-				jobs.push_back(createTexturedJob(description, revId, 3, grouping, true, texId));
-			}
-		}
+	jobs.push_back(createUntexturedJob(revId, 2, false, false));
```
FJThiel: Really minor, but I think it would be nice to have a small comment after each line stating the job description i.e. ``\\ Opaque, primitive 2, No normals``.

bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.cpp
```diff
@@ -44,130 +47,80 @@ static const size_t REPO_MODEL_LOW_CLUSTERING_RATIO = 0.2f;
...
+		// One cannot map a texture to a line, however, customers can assign materials with textures to lines
+		// so we need to be able to process them.
+		jobs.push_back(createTexturedJob(revId, 2, false, texId));
+		jobs.push_back(createTexturedJob(revId, 2, true, texId));
+
+		jobs.push_back(createTexturedJob(revId, 3, false, texId));
+		jobs.push_back(createTexturedJob(revId, 3, true, texId));
 	}
-
+
```
FJThiel: Rogue tab

bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.h
FJThiel: Ah, because they are all held by branchGroups, right?

FJThiel (changes requested):

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_nwd.cpp
sebjf: @FJThiel yes, or more precisely, Rvt or Ifc files that are embedded in/hosted by the NWD, as NWDs can contain both at the same time!

bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.h
sebjf: @FJThiel, yes indeed - keeping in mind premature optimisation is bad, I decided a shared ptr is quite heavy for this use case, as shared pointers are threadsafe, and we expect TransformInfo to be copied by value a lot. Since this is an private type I think it is reasonable it works with pointers into branchGroups, which will exist as long as the optimiser does.

bouncer/src/repo/manipulator/modelutility/repo_scene_manager.cpp
sebjf: Originally because I didn't want to change all the declarations in the tests - but then we did anyway! It is now part of the constructor!

bouncer/src/repo/manipulator/modeloptimizer/repo_optimizer_multipart.h
sebjf: Fixed!

bouncer/src/repo/core/model/bson/repo_bson_assets.h
sebjf: Fixed

FJThiel (approved):

AsiteSai (approved):

Issue #800, opened 2025-08-19
Remove federation support from bouncer


sebjf:
Federations are now handled by the backend. After they've operated this way for a release cycle, we should decommission the federation code in bouncer.

AsiteSai:
### Scope Analysis for Federation Code Removal

I have narrowed down the scope for removing federation-related code as follows.

---

### Client + Bouncer Library

The call chain for generating federations (**core bouncer functionality**)

```text
client/src/functions.cpp
  └─ generateFederation()
     └─ bouncer/src/repo/repo_controller.cpp
        └─ RepoController::createFederatedScene()
           └─ bouncer/src/repo/repo_controller_internal.cpp.inl
              └─ RepoControllerImpl::createFederatedScene()
                 └─ bouncer/src/repo/repo_manipulator.cpp
                    └─ RepoManipulator::createFederatedScene()
```

---

### Tests

Apart from the core functionality, there are federation related unit tests

```text
st_3drepobouncerClient.cpp
  TEST(RepoClientTest, CreateFedTest)
```

This^ test heavily relies on the following helper function:

```text
repo_test_utils.cpp
  produceCreateFedArgs()
```

The following file 
```text
repo_test_database_info.h
```
defines **five federation-related constants**. Some of these are reused by other unit tests (for example, **REPO_GTEST_DBNAME1_FED** is also used in the *getSceneFromReference* test), so I am currently considering leaving these unchanged.

---

### Bouncer Worker

The **bouncer_worker** Node.js tool also contains federation-related functionality:

```text
messageDecoder.js (entry point)
 - Parses federation commands ("genFed" switch/case)
    └─ jobQueueHandler.js
          - Invokes the bouncer CLI via createFed()
```

---

### Question Regarding createFed()

The **createFed()** method is also used when generating **stashes**, not just federations.

The flow looks like this:

```text
messageDecoder
  -> createFed()
     -> runBouncerCommand()
```

@sebjf  
From my understanding, **createFed()** in **jobQueueHandler.js** simply forwards the decoded message(regardless of whether the command is *Generate Federation* or *Generate Stash*) to **runBouncerCommand()**. So instead of deleting the **createFed()** function would it make more sense to rename it to

```text
runJob()
or
runCommand()
```
and would that fall under the scope of this task as well?

---

### Additional Question

Before I start deleting all this code, please let me know whether I should restrict the scope strictly to the **Bouncer Library**, or if I have missed any additional areas that should be considered.

[...]

PR #871, opened 2026-06-01
Issue #800 Remove federation support from bouncer

AsiteSai:
This fixes ISSUE [#800](https://github.com/3drepo/3drepobouncer/issues/800)

#### Description
- Removed Reference nodes that were earlier used in federations to reference scenes.
- Removed Federation related logic/comments/variables and tests.
- Removed JobQ functionality from bouncer worker.
- Throwing a more detailed message with an exception when we are unable to write files due to invalid fs permissions

#### Acceptance Criteria
- [x] As a Developer, I should not see any residual references to Federation, or Federation related dependencies in the Bouncer codebase.
- [x] As a Develpoer, I should see a slightly more helpful message when I don't have the right permissions set up and Bouncer Worker fails to upload files because of that.


AsiteSai (commented):
I reverted this change because error code 32 encompasses this issue. 

```
//Imported successfully, but failed to upload it to the database/fileshares
#define REPOERR_UPLOAD_FAILED 32
```

AsiteSai (commented):
I add the error check here, but I'm unable to propagate a custom error code from here as _uploadFileAndCommit()_ returns a bool.

bouncer/src/repo/core/model/repo_model_global.h
```diff
@@ -201,7 +201,6 @@
 #define REPO_NODE_TYPE_MATERIAL			"material"
 #define REPO_NODE_TYPE_MESH          	"mesh"
 #define REPO_NODE_TYPE_METADATA     	"meta"
-#define REPO_NODE_TYPE_REFERENCE		"ref"
```
sebjf, line 204: This definition we need to leave in place. This is because it will still exist in the database in older imports, so though we won't create new ones, its important to keep it 'reserved' so the string is not accidentally used for something else in the future!

sebjf (changes requested):
Looks great! Just one thing to put back below for documentation purposes.

bouncer/src/repo/core/model/repo_model_global.h
AsiteSai: Fixed in this [commit](https://github.com/3drepo/3drepobouncer/pull/871/commits/ab3cb4ffb90f694d9d848bf1a1fcbf6c75fc8524)

sebjf (approved):

PR #882, opened 2026-07-03
ISSUE #881 throw exceptions not pointers

### LINKED ISSUE #881: Regression bug in Navis importer
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/1004

sebjf:
This fixes #881 - an NWD importer regression that was showing up only in Docker.

The actual fix is in the `RepoMeshBuilder::face::face(std::initializer_list<repo::lib::RepoVector3D64> vertices, repo::lib::RepoVector3D64 normal)` constructor, which was not calling the overload that initialises `numUvs`, like the others.

Though in investigating this a couple of other minor issues were resolved and fixed:

1. RepoMeshBuilder no longer throws if already inside an exception handler, because that would lose the original exception.
2. RepoMeshBuilder throws objects instead of pointers, as throwing pointers is not correct.
3. CMAKE files that should have already been updated are committed.

There are no regression tests added because this is was not a logic error as such and so the only way to find such a bug would have been with valgrind, etc.


FJThiel (approved):
Looks good to me.

PR #895, opened 2026-07-30
Issue 768 - remove pdf support and add bouncer worker automated tests

### LINKED ISSUE #768: Add automated tests for bouncer_worker
bouncer_worker is not currently covered by unit tests on travis. This is annoying because we routinely get PRs from Snyk, and they can't be merged without someone checking them manually. We should have the worker covered by the automated tests so we can just accept these if travis passes.

sebjf:
This fixes #768

#### Description
1. Add automated tests for bouncer worker, so we can automatically check the dependency bot PRs from now on.
2. Remove PDF processing from bouncer worker, as this is no longer needed (and was breaking the tests).


sebjf:
@carmenfan requesting your review on this because its really more to do with devops/infrastructure than bouncer!

tools/bouncer_worker/tests/bouncerWorker.test.js
carmenfan, line 1: Can we put this under `tests/integration/`...

As this is a e2e integration test, it would be good to have the folder to indicate as such

tools/bouncer_worker/tests/tests.js
```diff
@@ -0,0 +1,359 @@
...
+const crypto = require('crypto');
+const fs = require('fs');
+const path = require('path');
+const { config, replaceSharedDirTag } = require('../src/lib/config');
+const { PROCESSING } = require('../src/constants/statuses');
+const { CLASH: CLASH_TYPE, IMPORT: IMPORT_TYPE, DRAWING: DRAWING_TYPE } = require('../src/constants/messageTypes');
+const QueueLabels = require('../src/constants/queueLabels');
+const queueLabels = require('../src/constants/queueLabels');
+
+const testClashQ = async () => {
```
carmenfan: There's quite a lot of repetition between the functions.

perhaps it would make sense to offload post and wait for callbacks to helper functions (with parameters for queue name, message content, callback function for consume etc) and reuse them?

tools/bouncer_worker/tests/tests.js
```diff
@@ -0,0 +1,359 @@
...
+	const clashq = config.rabbitmq.clash_queue;
+	const callbackq = config.rabbitmq.callback_queue;
+	const connection = await ampq.connect(config.rabbitmq.host);
+	const correlationId = crypto.randomUUID().toString();
+
+	// In practice, these are used to determine where to look for the clash run
+	// in the database. They are completely independent of which geometries
+	// are clashed, so can be anything here.
+
+	const project = 'testProject';
```
carmenfan: Unlikely to be important, but imo it's better practice to generate random strings for text so there's no chance of things only working because of fixed strings (highly unlikely, hence not important😆 )

tools/bouncer_worker/tests/tests.js
```diff
@@ -0,0 +1,359 @@
...
+const { PROCESSING } = require('../src/constants/statuses');
+const { CLASH: CLASH_TYPE, IMPORT: IMPORT_TYPE, DRAWING: DRAWING_TYPE } = require('../src/constants/messageTypes');
+const QueueLabels = require('../src/constants/queueLabels');
+const queueLabels = require('../src/constants/queueLabels');
+
+const testClashQ = async () => {
+	const clashq = config.rabbitmq.clash_queue;
+	const callbackq = config.rabbitmq.callback_queue;
+	const connection = await ampq.connect(config.rabbitmq.host);
+	const correlationId = crypto.randomUUID().toString();
```
carmenfan: better to have a `generateUUIDString()` function in helpers and reference that to reduce the amount of imports to crypto (not important now, but if this test library grows in the future)

tools/bouncer_worker/tests/tests.js
carmenfan, line 1: The file name is a bit confusing to me. When I looked at the structure of the folder for the changes, I see the .test.js, which is clear in convention what it does, but I tend not to understand why there's a tests.js without delving into the code andunderstandingd the contexts.

`bouncerWorker.test.js` is the main entry point and contains the actual definitions, but this is where thetest contents actually lives.

This is fine, but I feel like this file can do with better naming so it self-describes its intentions (e.g. something that conveys this is a list of reusable tests that talks to the queues and generates work items)

Alternatively you could just have all these in bouncerWorker.test.js, as you don't have another test file anyway (and refactor when/if needed) 🙈

tools/bouncer_worker/tests/tests.js
```diff
@@ -0,0 +1,359 @@
...
+
+	// In practice, these are used to determine where to look for the clash run
+	// in the database. They are completely independent of which geometries
+	// are clashed, so can be anything here.
+
+	const project = 'testProject';
+	const teamspace = 'testTeamspace';
+
+	const clashConfigDirectory = path.join(config.rabbitmq.sharedDir, correlationId);
```
carmenfan: Similar, maybe it's better to have a a setup helper function that generates a correlationId, creates the directary, takes in a list of files to copy into the folder and returns you the correctionId so all the functions can call

tools/bouncer_worker/tests/tests.js
```diff
@@ -0,0 +1,359 @@
...
+		await new Promise((resolve) => setTimeout(resolve, intervalMS));
+	}
+
+	assert.equal(matched, true);
+
+	await channel.close();
+	await connection.close();
+};
+
+module.exports = { testClashQ, testModelQ, testDrawingQ, testSingleQConsumer };
```
carmenfan: The convention we use in the source code and in 3drepo.io is to construct an object at the top of the file, declare local variables within functions, assign them to the object, and export the object.


This way, when you add a new function, you don't have to remember to go to the bottom and add it into your object (very easy to forget!) It's also very easy to tell which function is an internal helper and which is external facing.

carmenfan (changes requested):

tools/bouncer_worker/tests/unit/elastic.test.js
```diff
@@ -0,0 +1,238 @@
...
+		assert.deepEqual(calls.createIndex, ['io-bouncer']);
+		assert.equal(calls.putMapping.length, 1);
+		assert.equal(calls.putMapping[0].index, 'io-bouncer');
+		assert.equal(calls.putMapping[0].body.properties.Owner.type, 'keyword');
+		assert.equal(calls.createDoc.length, 1);
+		assert.equal(calls.createDoc[0].index, 'io-bouncer');
+		assert.equal(calls.createDoc[0].id, 'hashed-id');
+		assert.equal(calls.createDoc[0].refresh, true);
+		assert.equal(calls.createDoc[0].body.namespace, 'worker-a');
+		assert.equal(logs.info.some(([msg]) => msg.includes('Succesfully connected to elastic-cloud-id')), true);
```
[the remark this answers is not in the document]
carmenfan: Maybe we should just fix the typo 🙈

tools/bouncer_worker/tests/unit/bouncerClient.test.js
```diff
@@ -0,0 +1,195 @@
...
+			}
+			return 0;
+		},
+	};
+
+	const bouncerClient = moduleRequire(bouncerClientPath);
+	return { bouncerClient, calls };
+};
+
+describe('bouncerClient.js', () => {
```
carmenfan: Would be better to extract this from __filename (nodejs variable for the path of the file the line of code resides in).

So in the event if we ever renamed the file, no one would have to remember to update this string.

tools/bouncer_worker/tests/unit/bouncerClient.test.js
```diff
@@ -0,0 +1,195 @@
...
+	test('testClient logs status, sets env and runs bouncer test command', async () => {
```
carmenfan: As they are unit tests (we're testing the functions separately, it would be better if we separate them with different `describe` to separate each function.

For example, the convention we took in 3drepo.io is to have an outer describe for the src file (https://github.com/3drepo/3drepo.io/blob/master/backend/tests/v5/unit/utils/webRequests.test.js#L150), then all it does is do any global beforeAll and AfterAll, and call individual functions that correspond to the function it will test. In that test, the function will then have its [describe to group the individual test cases for that function](https://github.com/3drepo/3drepo.io/blob/master/backend/tests/v5/unit/utils/webRequests.test.js#L57). That makes the structure of the test file much more deterministic and easier to navigate

tools/bouncer_worker/tests/unit/bouncerClient.test.js
```diff
@@ -0,0 +1,195 @@
...
+		assert.equal(calls.loggerInfo.some(([msg]) => msg.includes('Checking status of client...')), true);
+		assert.equal(calls.loggerInfo.some(([msg]) => msg.includes('Machine Instance ID is set to instance-abc')), true);
+		assert.equal(calls.loggerInfo.some(([msg]) => msg.includes('Bouncer call passed')), true);
+
+		assert.equal(calls.run.length, 1);
+		assert.equal(calls.run[0][0], expectedBouncerPath);
+		assert.deepEqual(calls.run[0][1], ['/tmp/bouncer-config.json', 'test']);
+		assert.deepEqual(calls.run[0][2], { logLabel: { label: 'INIT' } });
+
+		assert.equal(process.env.BC_ENV_ALPHA, 'alpha');
```
carmenfan: it would be better to just generate random strings for both the env variable names and the values they hold and reference them 

Same with the rest of the test cases in this file (and also other test files we're hard coding data values when it's not needed)

tools/bouncer_worker/tests/unit/config.test.js
```diff
@@ -0,0 +1,287 @@
...
+	console.log = originalConsoleLog;
+	process.umask = originalUmask;
+	if (originalEnvConfig === undefined) {
+		delete process.env.BOUNCER_CONFIG;
+	} else {
+		process.env.BOUNCER_CONFIG = originalEnvConfig;
+	}
+};
+
+afterEach(cleanup);
```
carmenfan: Not super familiar with Node's built-in test convention (we use Jest as Node's test runner didn't exist when we built 3drepo.io). If it's the same as Jest, then it would be good to put this `afterEach` inside your describe group (i.e. around L127) , so it's clear from the test declaration what tests you're running and what pre/post processing happens.

At the moment, the intention of the scope isn't super clear (esp if you end up with some pre/post processing for a sub-group of tests within a file in the future), and anyone maintaining this can easily miss this call.

carmenfan (dismissed):

carmenfan (approved):

PR #901, opened 2026-08-18
Support Civils3D proxy objects

### LINKED ISSUE #868: Support Civils3D proxy objects
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/973

Civils3D has an option to enable proxy objects when saving DWGs. We should read this geometry and as much metadata as we can get a hold of.

### Goals

- Proxy entities render using their stored graphics instead of being dropped or shown as placeholders
- Civil3D TIN surfaces render as real triangulated terrain geometry, not just a bounding box
- Civil3D and Plant3D proxies carry identifying metadata (class, application, XData, extension-dictionary properties) even when geometry fidelity is limited
- Civil3D and Plant3D support are independently addable/removable, so a future app (or a Plant3D-specific geometry type) can be added without reworking the core DWG import path
- Import diagnostics report proxy/Civil3D/Plant3D counts to make regressions visible on future drawings

### Tasks
- [x] Detect proxy entities and classify by originating application (Civil3D / Plant3D / custom)
- [x] Consolidate proxy inspection into a single pass per entity, removing redundant ODA casts and database queries
- [x] Implement stored-graphics replay fallback rendering for proxies (worldDraw)
- [x] Implement Civil3D TIN surface geometry capture — triangle/edge extraction, duplicate-edge filtering, degenerate-triangle rejection, material handling
- [x] Compute TIN surface metadata (rendered point count, min/max elevation, name derived from layer)
- [x] Extract Civil3D extension-dictionary stored properties (Station, Offset, Elevation, Grade)
- [x] Extract general/XData/Common-Data-Access properties for proxy entities
- [x] Add import diagnostics (entity/geometry/proxy/Civil3D/Plant3D counters, summary log)
- [x] Make XData and extension-dictionary reads lazy to avoid unnecessary database opens on proxies that don't need them
- [x] Extract Civil3D/Plant3D-specific logic behind a shared interface so either can be added or removed independently of the core importer
- [x] Verify the decoupling: remove one app's handler and confirm no changes are needed outside its own files
- [x] Modify/Add required unit test
- [x] QA: import a real DWG with Civil3D proxies; confirm geometry and metadata match expected output — 0.5 day (outstanding)

### Related Resources

MockUp : `link to mockup`

vmavadiya:
This fixes #868

#### Description
In this PR we have support for Civil3D proxy entity object rendering and metadata for that object will be displayed in attribute data of 3drepo platform.

#### Acceptance Criteria
- [ ] Check normal dwg file it should render correctly.
- [ ] Check Civil3D dwg file without proxygraphics=1, It should not render proxy object.
- [ ] Check Civil3D dwg file with proxygraphics=1, It should render proxy object.


bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/civil3d_proxy_handler.cpp
```diff
@@ -0,0 +1,735 @@
...
+{
+	if (!hasFaceMaterial)
+	{
+		faceMaterial = collector->getLastMaterial();
+		hasFaceMaterial = true;
+	}
+}
+
+bool Civil3DProxyHandler::TinCapture::addTriangle(
+	GeometryCollector* collector,
```
sebjf: Is there a reason to pass the collector in to each method as opposed to making it a member of `Civil3DProxyHandler` and initialising it when the handler is created? `GeometryCollector` is created once per import, so it should not change.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/civil3d_proxy_handler.cpp
```diff
@@ -0,0 +1,735 @@
...
+	if (samePoint(p0, p1) || samePoint(p1, p2) || samePoint(p2, p0)) return false;
+
+	GeometryCollector::Face triangle;
+	triangle.push_back(p0);
+	triangle.push_back(p1);
+	triangle.push_back(p2);
+	capturedTriangles.push_back(triangle);
+	captureMaterialIfNeeded(collector);
+
+	addEdge(collector, p0, p1);
```
sebjf: Do you recall if the standard DWG importer defaults to `shaded with edges`? If not, I think we can skip adding edges for each triangle. From discussions around the Revit importer it seems most customers prefer not to have explicit edges. We could also make this optional, using the `style` parameter of the config.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/civil3d_proxy_handler.cpp
```diff
@@ -0,0 +1,735 @@
...
+	collector->setMaterial(edgeMaterial);
+	collector->addFace({ p0, p1 });
+	return true;
+}
+
+bool Civil3DProxyHandler::TinCapture::addPolyline(
+	GeometryCollector* collector,
+	const std::vector<repo::lib::RepoVector3D64>& points)
+{
+	if (points.size() != 4) return false;
```
sebjf: This method is a little confusing, because because its called addPolyline, but it is creating a triangle (and skips any lines that cannot be turned into triangles!). We should either create the polyline using edges, or call it something else.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/civil3d_proxy_handler.cpp
```diff
@@ -0,0 +1,735 @@
...
+	GeometryCollector* collector,
+	const std::string& parentLayerId,
+	const std::string& sourceEntityId) const
+{
+	for (size_t i = 0; i < capturedTriangles.size(); ++i)
+	{
+		auto faceLayerId = sourceEntityId + ":TIN_FACE:" + std::to_string(i);
+
+		auto faceContext = collector->makeNewDrawContext();
+		collector->pushDrawContext(faceContext.get());
```
sebjf: It looks like this is creating a new context & layer for each triangle in `capturedTriangles`. It doesn't look like the  material can change inside the loop, so would it not be better to create one context/layer?

Contexts can be quite expensive so I don't think we want one per triangle!

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/civil3d_proxy_handler.h
```diff
@@ -0,0 +1,114 @@
...
+#include "geometry_collector.h"
+#include "repo/lib/datastructure/repo_structs.h"
+
+namespace repo {
+	namespace manipulator {
+		namespace modelconvertor {
+			namespace odaHelper {
+				/* Civil3D-specific proxy handling: class-name/XData classification,
+				extension-dictionary stored properties, the Civil3D proxy-class
+				display-name map, and TIN surface geometry capture. This is the only
```
sebjf: Do we know what `TIN` is an acronym for? Would be good to have it in the comments the first time it is used!

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/CMakeLists.txt
```diff
@@ -7,18 +7,21 @@
...
 		${CMAKE_CURRENT_SOURCE_DIR}/data_processor_rvt.cpp
+		${CMAKE_CURRENT_SOURCE_DIR}/dwg_proxy_inspector.cpp
 		${CMAKE_CURRENT_SOURCE_DIR}/file_processor.cpp
 		${CMAKE_CURRENT_SOURCE_DIR}/file_processor_dgn.cpp
 		${CMAKE_CURRENT_SOURCE_DIR}/file_processor_dwg.cpp
 		${CMAKE_CURRENT_SOURCE_DIR}/file_processor_nwd.cpp
 		${CMAKE_CURRENT_SOURCE_DIR}/file_processor_rvt.cpp
 		${CMAKE_CURRENT_SOURCE_DIR}/geometry_collector.cpp
 		${CMAKE_CURRENT_SOURCE_DIR}/helper_functions.cpp
+		${CMAKE_CURRENT_SOURCE_DIR}/plant3d_proxy_handler.cpp
```
sebjf: Are the plant3d files meant to be in this PR?

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor.h
```diff
@@ -79,6 +79,18 @@ namespace repo {
 						MaterialColours& matColors,
 						repo::lib::repo_material_t& material);
 
+					virtual void processTriangleOut(
```
sebjf: Instead of adding new methods, if we want `DataProcessorDwg` to be able to override the implementations, we can instead make the original `triangleOut`, etc methods protected instead of private, and have subclasses override them directly, and in those overrides call base:: as needed.

I think this would be easier to read, as having two methods suggests there is a functional difference between them (e.g. _process_ has additional logic), but in practice it is a wrapper.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_dwg.h
```diff
@@ -51,7 +61,56 @@ namespace repo {
...
+					void processPolylineOut(
+						OdInt32 numPoints,
+						const OdGePoint3d* vertexList) override;
+
+					void polygonOut(
+						OdInt32 numPoints,
+						const OdGePoint3d* vertexList,
+						const OdGeVector3d* pNormal = 0) override;
+
+					void shellProc(
```
sebjf: For these new methods such as `shellProc`...

`DwgDataProcessor` is a subclass of `OdGiGeometrySimplifier` (https://docs.opendesign.com/tv_api_cpp/OdGiGeometrySimplifier__polygonOut@OdInt32@const_OdGePoint3d_@const_OdGeVector3d_.html).

This means the base class will call subsequently lower level methods by default until it gets down to those that have no default implementation. The documentation will say which one is which. For example, following the docs for [shellProc](https://docs.opendesign.com/tv_api_cpp/OdGiGeometrySimplifier__shellProc@OdInt32@const_OdGePoint3d_@OdInt32@const_OdInt32_@const_OdGiEdgeData_@const_OdGiFaceData_@const_OdGiVertexData_.html), it goes:

shellProc -> generateShellFaces -> shellFaceOut -> triangleOut -> polygonOut 

Only when it gets to polygonOut, does it say explicitly that clients should override it: https://docs.opendesign.com/tv_api_cpp/OdGiGeometrySimplifier__polygonOut@OdInt32@const_OdGePoint3d_@const_OdGeVector3d_.html

If we instead put the proxy capture logic in `polygonOut` by making it possible for `DwgDataProcessor` to override it, then we can forgo all these overrides.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/dwg_proxy_inspector.cpp
```diff
@@ -0,0 +1,478 @@
...
+	}
+
+	// No app detected yet for this file: probe each app's class-name rule
+	// without constructing an instance, and construct exactly the one that
+	// matches - a DWG is authored by one app, never both.
+	if (Civil3DProxyHandler::matchesClassName(originalClass))
+	{
+		activeHandler = std::make_unique<Civil3DProxyHandler>();
+	}
+	else if (Plant3DProxyHandler::matchesClassName(originalClass))
```
sebjf: Was it intentional to have Plant3D stuff here? It is perfectly OK to have it in this PR, I am just thinking when you do the Plant3D PR it might conflict!

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/file_processor_dwg.cpp
```diff
@@ -217,9 +225,32 @@ uint8_t FileProcessorDwg::readFile()
...
+		}
 
 		OdDbDatabasePtr pDb = svcs.readFile(getFilename());
 
+		// ===== ADDED: Configure database to use proxy graphics =====
+		// PROXYGRAPHICS system variable controls how custom objects display:
+		// 0 = No proxy graphics (bounding box only)
+		// 1 = Show proxy graphics
+		try {
+			//pDb->setPROXYGRAPHICS(1); // Enable proxy graphics display
```
sebjf: I believe we discussed this before and what this flag means is that the app (in this case ODA, but in the original case AutoCAD) would *write* the proxy geometry. Therefore we cannot just set this at import time and get proxy objects. Is that right? Or if we set this flag before drawing the file we get the proxy objects? Either way I think we should put whatever the answer is in a comment!

sebjf (changes requested):
Looks good for the most part! There are a few small comments in-line. The main comment is around the geometry proxy functionality, as this bit I found quite hard to follow to start with.

A number of the comments talk about replay/storage. it looks like the Proxy Handler is capturing triangles and edges to `capturedTriangles`, however, it is also adding them to the collector immediately as well. Also, the `addEdge`, `addTriangle` etc methods are in `Civil3DProxyHandler` but are actually called by the DataProcessor overrides. Further in doDraw we have a flag, `replayStoredProxyGraphics`, but I don't actually see where the geometry is being replayed. It looks like the `ProxyAppHandler` returns a `ProxyGeometryCapture` - i.e. it is per-app, not per-entity. I do see the `applyFaceLayers` call, but then it is followed by `clearTriangles`.

From the code, and other comments, it looks like rather than the app handler wanting to capture & replay geometry, instead it wants to change the behaviour of `GeometryCollector`, so that triangles are rendered as wireframes, edges have a specific material, faces are given their own layers, etc.

Have I understood this correctly or is there something I am missing?

If the goal is indeed that the proxy handler wants different behaviour to `DataProcessor`/`GeometryCollector`, then I think we can simplify this a bit, as especially the conditional logic at the end of doDraw is not always obvious in its intention.

Though, looking at `Civil3DProxyHandler` I am not sure if we even need a separate handler here? We don't want each face to be in a separate mesh for performance reasons, as per the in-line comments. In previous discussions around the Revit importer, customers preferred not to use edge shading, so if we are creating solid surfaces, we don't necessarily need lines as well for each face edge. If we did want this behaviour, it might be more straightforward to implement it with a flag on the `DataProcessorDwg` level, controlled by the style option of the config. Then it would also be available to other entities.

The other thing I see it do is compute additional metadata, however we could get this from the bounds from the ctx object afterwards.

Have a look at the comments and let me know what you think!

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/file_processor_dwg.cpp
vmavadiya: i was forget to removed it, i will do it because we don't have ProxyModule in ODA so we can't set PROXYGRAPHICS=1 from code itself, It always from AutoCAD

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/dwg_proxy_inspector.cpp
vmavadiya: I have just put placeholder for now, As i am restructuring i will removed it

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/CMakeLists.txt
vmavadiya: It will removed. i was forgot too removed it

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/civil3d_proxy_handler.h
vmavadiya: The **TIN** stands for **Triangulated Irregular Network** i will update it

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/civil3d_proxy_handler.cpp
vmavadiya: As you suggest i will try too check with your flow mention in https://github.com/3drepo/3drepobouncer/pull/901#pullrequestreview-4970346244 after that it will become invalid

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/civil3d_proxy_handler.cpp
vmavadiya: This will removed from per triangle and manage it with per layer

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor.h
vmavadiya: i have make it protected in base class and override it in derived class which are required only

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_dwg.h
vmavadiya: I have remove all the method which introduced for POC and only override for now polygonOut method in DwgDataProcessor

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/dwg_proxy_utils.h
```diff
@@ -0,0 +1,104 @@
...
+					// between adjacent triangles is only added to the wireframe overlay once.
+					// Cleared at the start of each entity's draw.
+					std::unordered_set<std::string> currentSurfaceEdgeKeys;
+
+					// Per-entity dedup of triangle vertices seen for the same reason (keyed by
+					// pointKey(p)). Its size is used as the surface's "Number Of Points"
+					// metadata - an approximation, since Civil3D's true TIN point count isn't
+					// otherwise available from the tessellated proxy graphics. Cleared at the
+					// start of each entity's draw.
+					std::unordered_set<std::string> currentSurfacePointKeys;
```
sebjf: I think we can do without this, as I can't imagine customers needing this info, and also, it is sort of an implementation detail since it is based on how we do the edge extraction (in fact, because of the de-dupe of MeshNode, it may not even be accurate by the time it gets into viewer!)

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/dwg_proxy_utils.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+	info.graphicsType = info.entity->graphicsMetafileType();
+
+	try { info.graphicsPE = OdDbEntityWithGrDataPE::cast(entity); }
+	catch (...) {}
+
+	return info;
+}
+
+bool DwgProxyUtils::drawStoredProxyGraphics(OdDbEntityPtr pEntity, const ProxyInfo& info, OdGiWorldDraw* worldDraw)
```
sebjf: I think we should name this something like "drawProxyGraphics", because the word "stored" suggests something that has been previously initialised, but that is not the case (at least for our code...). One can call this with any pEntity and worldDraw and it will render directly into the context.

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/helper_functions.cpp
```diff
@@ -65,6 +65,30 @@ int repo::manipulator::modelconvertor::odaHelper::compare(double d1, double d2)
...
+{
+	// Quantise to the same tolerance samePoint()/compare() use, so points that
+	// compare equal always produce the same key.
+	const double keyScale = 1.0 / DOUBLE_TOLERANCE;
+	return std::to_string(static_cast<long long>(p.x * keyScale)) + "," +
+		std::to_string(static_cast<long long>(p.y * keyScale)) + "," +
+		std::to_string(static_cast<long long>(p.z * keyScale));
+}
+
+std::string repo::manipulator::modelconvertor::odaHelper::edgeKey(const repo::lib::RepoVector3D64& a, const repo::lib::RepoVector3D64& b)
```
sebjf: For performing de-duplication, I would not use strings here, because they are quite expensive for this use case. 

Each string will allocate its own internal buffer to go along with the memory on the stack, and to represent each vertex it would consume ~60 bytes (whereas each vertex is 12 bytes natively). And we have that 60 bytes for each element of each pair, because there is no indexing.

It is a good question where to do the deduplication. 

Really, it should be in the Geometry Collector or its context, but those classes are quite old and it would be a pain to refactor them, out of scope for this I'd say.

I think we should keep it here but make a couple of adjustments:

(1) Create a dedicated struct for the set key type instead of using a string. (E.g. `EdgeKey`)

This type will need a hash function and an equality comparison operator to be used with one of the sets.

For hashing, you can use the existing helper (https://github.com/3drepo/3drepobouncer/blob/master/bouncer/src/repo/lib/repo_hash_combine.h).

Here is an example of it hashing one vector: https://github.com/3drepo/3drepobouncer/blob/35e9d7a7c83a319c45753ac500a4cf3b2640f225/bouncer/src/repo/manipulator/modelutility/clashdetection/geometry_utils.cpp#L44

Just like you do now, I would hash each vertex individually, then order them based on the per-vertex hash values, then the hash of the edge is the combination of each of those hashes (in order).

For comparison, you can just do an equality comparison between each of the (already ordered) vectors (that operator exists for repovector).

(2) Consider using hopscotch set instead of std::unordered_set, as it will be more efficient. It is already in bouncer: https://github.com/3drepo/3drepobouncer/blob/master/bouncer/src/repo/lib/hopscotch_map/hopscotch_set.h

All of this can live in dwg_proxy_utils.cpp. Instead of exposing the `currentSurfaceEdgeKeys` and `edgeKey` etc, you could have a couple of methods on `ProxyInfo`, e.g. `HasEdge()` and `ResetEdges()`. `HasEdge` returns false the first time the edge is encountered & adds it to the map, which is now a private member. `ResetEdges` does what the clear commands do currently.

sebjf (changes requested):
Two very minor comments and a suggestion for the de-duplication, then I think we are good!

bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/helper_functions.cpp
vmavadiya: The changes are done as you mention in https://github.com/3drepo/3drepobouncer/pull/901#discussion_r3950597143

sebjf (approved):

PR #905, opened 2026-08-21
Issue 904 - clean up existing artefacts when importing to a specific revision id

### LINKED ISSUE #904: Delete any existing nodes from the .scene collection matching the revId if provided
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/1023

### Goals

- [ ] When a revId is provided, any existing import artefacts should be deleted.

### Tasks
- [ ] Check collection for existing nodes with revision id
- [ ] Delete nodes, collecting ref nodes if they have binaries
- [ ] Delete ref nodes and linked files from disk

### Related Resources

MockUp : `link to mockup`

sebjf:
This fixes #904

#### Description

This PR adds a step to the model import that cleans up any existing scene nodes under the revision id that is being imported to. The goal is to remove artefacts of previous imports. Regardless of the state of these imports, any overlap within a revision will confuse the streaming repobundle generation, so they must be removed.

To support this,

1. `deleteFileAndRef` was updated so it can delete multiple files & ref nodes at once, instead of just one, as we can expect the clean-up step to want to delete large numbers of ref nodes & doing it one by one would be inefficient.
2. The Or query variadic constructor will no longer accept single arguments - when used with the RepoQuery variant this was too permissive and hiding coding errors that should have shown up at compile time.
3. The database handler can drop documents by criteria, instead of having to read them back first - again so we can clear an entire import in one go.
4. A system test was added for this, along with the required supporting methods: those to create orphaned artefacts, write a model config (so we can import to the same rev id over again), and check the integrity of the imports afterwards.
5. RepoScene was updated so addNodes matches the documentation, and the commitScene unit test explicitly tests only the legacy scene behaviour (i.e. where RepoScene is built fully in-memory).

#### Acceptance Criteria

- [x] When importing, existing nodes for the revision id under the scene collection are deleted
- [x] Both orphaned nodes and existing root nodes are deleted
- [x] Any ref nodes referenced by these are also deleted (ref nodes cannot cross process lifetime boundaries when importing)
- [x] Any linked files of those ref nodes are deleted
- [x] Existing imports (different revision ids) are unaffected
- [x] If there are no nodes under the revision id, the clean-up does nothing


bouncer/src/repo/core/handler/fileservice/repo_file_manager.h
```diff
@@ -104,9 +105,9 @@ namespace repo {
 					 * Delete file ref and associated file from database.
```
FJThiel:
```suggestion
					 * Delete file refs and associated files from database.
```

test/src/unit/repo_test_utils.cpp
```diff
@@ -403,4 +411,146 @@ void testing::unsetupTextures()
...
 #endif // WIN32
+}
+
+void testing::writeImportConfig(
+	const std::string& importFilePath,
+	const repo::manipulator::modelconvertor::ModelImportConfig& config,
+	const std::string& configPath
+)
+{
+	boost::property_tree::ptree jsonTree;
```
FJThiel: Didn't we want to move away from the boost json in favour of RapidJSON?

sebjf: Good point! Replaced!

FJThiel (approved):

PR #910, opened 2026-09-07
Issue 909 - Update DWG Entity Handle behaviour

### LINKED ISSUE #909: Make DWG entity handles consistent across file formats
### Description
Product issue: https://github.com/3drepo/3D-Repo-Product-Team/issues/388

We would like to support DWG Entity Handles as External Ids. To do this, we need to ensure they are consistent when imported via NWDs as well as Native Imports, while also not breaking existing functionality.

The main challenge here is NWDs, as the NWD tree will often include composite parts in a more detailed breakdown than is presented in the 3DR tree. We need to ensure then that metadata (i.e. unique Ids) for DWGs, RVT and IFC files is attached to the correct node. Currently, the importer inherits parent metadata for NWD nodes, which works for RVT and IFC files, but not for DWGs.

### Goals

- [ ] The Element ID of the native BIM element must always appear on the Leaf Node of the 3DR tree.
- [ ] The Element ID/Entity Handle/IFC GUID of this node must match that of the Element in the native tool.
- [ ] For Navis files, the node in the Navis Tree containing the native BIM Element Id must correspond to the Leaf Node in the 3DR tree (which also contains that ID).
- [ ] The Leaf Node must encompass all geometry below it.
- [ ] Element Ids must be consistent across revisions.

### Tasks
- [x] Create or extend unit tests to check the tree of nestedBlocks.dwg and groupsAndReferences.dwg
- [x] Create or extend unit tests to check the tree of floors_and_levels.rvt
- [x] Create or extend unit tests to check the tree of floors_and_levels.ifc
- [x] Create unit tests to check the consistency of Nwds hosting all of the above.
- [x] Remove brackets from DWG handles
- [x] Make sure Entity Handles in Navis hosted DWGs are represented as hex strings.

### Related Resources

MockUp : `link to mockup`

sebjf:
This fixes #909

#### Description
This PR makes some small changes to improve the consistency of DWG Entity Handles across formats, so they can be used as external Ids.

1. The array decorators are removed from the native DWG importer, and in the Nwd importer the handle is converted to a hexadecimal string, so that the handles have the same representation between Dwgs, Nwds, the plug-ins, and native tools (AutoCAD).
2. The metadata inheritance in the Nwd has been 'inverted' so that instead of the parent overriding the child entries, the child entries pull only missing metadata from the parent. This keeps the behaviour where a leaf node inherits, e.g. the Type block from the parent, but avoids the situation where a parent node can override the Id of its children. This case was actually very rare before, because Revit and Ifc files when imported into Navis typically don't have BIM element instances as direct children of other element instances, so for these files the behaviour for their Ids is unaffected by this change.

Finally, the unit tests have been updated to cover the expected behaviour around ids and tree for all these file types and their combinations.

### DIFF
--- a/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_nwd.cpp
+++ b/bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_nwd.cpp
@@ -94,6 +94,7 @@ static std::vector<std::string> ignoredCategories = { "Material", "Geometry", "A
 static std::vector<std::string> ignoredKeys = { "Item::Source File Name" };
 static std::vector<std::string> keysForPathSanitation = { "Item::Source File", "Item::File Name" };
 static std::string sElementIdKey = "Element ID::Value";
+static std::string dwgEntityHandleKey = "Entity Handle::Value";
 
 // The Vectorizer and Simplifier are used to tessellate objects that are not
 // stored as meshes in the Nwd, such as text.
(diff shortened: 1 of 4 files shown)


bouncer/src/repo/manipulator/modelconvertor/import/odaHelper/data_processor_nwd.cpp
```diff
@@ -187,6 +188,49 @@ repo::lib::RepoVector2D convertPoint(OdGePoint2d pnt)
...
+			stream << std::uppercase << std::hex << (int64_t)d;
+			return stream.str();
+		}
+
+		std::string operator()(const std::string& s) const {
+			return s;
+		}
+
+		std::string operator()(const tm& t) const {
+			throw new repo::lib::RepoException("Cannot convert boolean to hex string");
```
FJThiel:
```suggestion
			throw new repo::lib::RepoException("Cannot convert time struct to hex string");
```

FJThiel (approved):

